PR #187, opened 2015-01-01
Add testcase for org.ethereum.vm.LogInfo

marcegarba:
Added a simple test_3() method to test class org.ethereum.vm.LogInfoTest.


romanman:
Good start.

PR #188, opened 2015-01-01
use javax.inject instead of autowired

ligi:
background: http://stackoverflow.com/questions/7142622/what-is-the-difference-between-inject-and-autowired-in-spring-framework-which/7144137#7144137

### DIFF
--- a/ethereumj-core/build.gradle
+++ b/ethereumj-core/build.gradle
@@ -65,6 +65,7 @@ dependencies {
     compile "org.hibernate:hibernate-core:${hibernateVersion}"
     compile "org.hibernate:hibernate-entitymanager:${hibernateVersion}"
     compile "commons-dbcp:commons-dbcp:1.4"
+    compile "javax.inject:javax.inject:1"
     compile("com.googlecode.json-simple:json-simple:1.1.1") {
         exclude group:'junit', module:'junit'
     }
--- a/ethereumj-core/src/main/java/org/ethereum/core/Account.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Account.java
@@ -4,7 +4,7 @@ import org.ethereum.crypto.ECKey;
 import org.ethereum.manager.WorldManager;
 import org.ethereum.util.Utils;
 
-import org.springframework.beans.factory.annotation.Autowired;
+import javax.inject.Inject;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
@@ -28,7 +28,7 @@ public class Account {
     private Set<Transaction> pendingTransactions =
             Collections.synchronizedSet(new HashSet<Transaction>());
 
-    @Autowired
+    @Inject
     WorldManager worldManager;
 
     public Account() {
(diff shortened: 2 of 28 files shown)


cbeams:
@ligi, the DEV build breakage above was due to unused JDK 8 stream API imports still hanging around. I've just removed those in commit 6c559e6.

As for favoring the use of `@Inject` over `@Autowired`, that is fine in principle, but the larger question is whether we should be annotating these types at all. Typically, injection annotations are applied to application-level code, not library code. From the point of view of applications depending on ethereumj, ethereumj should probably be thought of and treated as a typical third-party dependency, i.e. where the wiring of ethereumj types is handled (in the Spring case) by `@Bean` methods within a `@Configuration` class, or (in the Guice case) as bindings within a `Module`, or simply by hand (if one is not using a DI framework).

I'll talk with @romanman about the above.

Also, @ligi, assuming you're using IDEA, note that there is a shared configuration for IDEA's import optimization now present in `.idea/codeStyleSettings.xml`. This should be picked up automatically, meaning that you can run "optimize imports" prior to committing and conform to the conventions there.

romanman:
@cbeams @ligi : 

The etherj lib started as non IoC application 
while I tried to keep everything simple and 
straight forward but at some point , I took 
it there by identifying certain modules that 
are very well defined like BlockChain and 
Repository and the main purpose is to 
be able to use injection in testing.

@ligi : I don't have any opinion on Inject vs Autowired
also as far as I know functionally  they are the same 
but I have some problem to auto merge your PR is it 
possible for you to merge it locally with recent changes 
and to file it again.

cbeams:
@romanman note that submitters of pull requests can rebase, merge or add new commits against the branch they submitted a pull request from, and those changes will automatically be reflected here in the GitHub pull request UI. So there's no need to close a PR and file a new one just because there are merge issues.

Reopen this if you like, but from my side, there's little value to switching to `@Inject` in our case, given that other Spring annotations that have no `javax.inject.*` equivalents are already in use (such as `@ComponentScan` and `@EnableTransactionManagement`). If upon review we want to keep injection annotations in the library types themselves (which I'm still doubtful about... need to spend more time with this), then it may make sense to do a more complete overhaul and try to eliminate all Spring injection annotations in favor of their `javax.inject.*` equivalents (and stop using those that have no equivalent). In any case, it would certainly be good to avoid forcing ethereumj users to depend on Spring if they don't otherwise want to. But again, in the meantime, I don't see much value in porting to `javax.inject` only partially in the meantime.

Also for the record, `@Inject` and `@Autowired` are for the most part equivalent, but not exactly so. They are treated equally throughout the Spring container, except as regards `@Autowired`'s `#required` attribute, which `@Inject` has no equivalent for. This is a moot point for ethereumj, though, as it does not use `@Autowired#required` anywhere.

romanman:
@cbeams @ligi : 

Yeah I guess it would be reasonable to move to `@Inject` if 
we would have any thoughts to move to other IoC vendor
but we will 100% confident about spring so I guess will 
stay with `@Autowired`.

cbeams:
> Yeah I guess it would be reasonable to move to @Inject if 
> we would have any thoughts to move to other IoC vendor
> but we will 100% confident about spring so I guess will 
> stay with @Autowired.

Well, it's not really about us, right? It's about other users of etherumj and what *they* prefer for DI. It's completely reasonable to imagine a future Android-based built on ethereumj that uses Dagger for DI. They wouldn't appreciate dragging in a Spring dependency.

We can probably just leave this alone for the moment, but something to take another look at prior to 1.0 in any case.

PR #189, opened 2015-01-04
Add tests for Messages in p2p/eth

faizkhan00:
This includes a few new tests:

DisconnectMessageTest.java - Added bounds checking
GetPeersMessageTest.java - broke this out into its own file for testing
HelloMessageTest.java - Added constructor, fail/bounds test
PeerTest.java - test peer object
StatusMessageTest.java - Added constructor, fail/bounds test

I'll have to run things with the new gradle framework, but these should all run and pass as well

### DIFF
--- a/ethereumj-core/src/test/java/test/ethereum/net/DisconnectMessageTest.java
+++ b/ethereumj-core/src/test/java/test/ethereum/net/DisconnectMessageTest.java
@@ -1,5 +1,11 @@
 package test.ethereum.net;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+
+import java.lang.System;
+
 import org.ethereum.net.message.ReasonCode;
 import org.ethereum.net.p2p.DisconnectMessage;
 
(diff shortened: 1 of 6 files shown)


faizkhan00:
Want to say : awesome job @cbeams and co. for putting in the gradle build system, committed a fix based on the CI build

Sadly only a 0.37% increase in coverage

faizkhan00:
Should I worry about this: http://build.ethdev.com/builders/Linux%20EthereumJ%20PRs/builds/29 ?

build gradlew failed 
install install failed 

the logs suggest this is due to a javadoc test failing, and nothing related to this commit

cbeams:
Thanks, @faizkhan00. For future reference, please see the follow-up commits in f34d1f4. We'll put together a `CONTRIBUTING.md` that makes this more explicit.

faizkhan00:
Works for me, I appreciate that @cbeams for making those changes and have updated my .vimrc accordingly. There were some comments left in the pre-cleaned tests which was deliberate- as a debugging reminder (especially in the null cases, which if toString() isn't commented will throw an exception), just wondering what rules apply there?

cbeams:
> just wondering what rules apply there

What rules apply to leaving comments in tests, you mean?

faizkhan00:
"What rules apply to leaving comments in tests, you mean?"

Yes, I realize now I might have just left a few words to describe why I left it. I won't miss those comments now that they've gone, but would that satisfy in the case of leaving commented out code ?

cbeams:
Generally, commented-out code is to be avoided. If there's a special case for it, just make it clear with a comment.

faizkhan00:
Okay, will do.

PR #196, opened 2015-01-10
dry and clean precompiled contracts

ligi:


romanman:
Sorry @ligi it's an unnecessary complication
I prefer to keep it simple and that generalization
doesn't help here. 

But thanks for going over the code another 
pair of eyes can be always useful and we 
are working on a certain plan to reward our 
volunteers for certain task around this project
so don't worry you will have the chance to make
changes on the code.

ligi:
hm - it is not about generalization - it is about DRY code - it makes it easier to read ( gives semantic meaning ) - reduces the amount of code and makes it less error-prone 
http://en.wikipedia.org/wiki/Don%27t_repeat_yourself
Perhaps you give it a second look ( you might have miss-read the first PR in the morning ) - I will continue to look over the code anyway because I think it will become an important foundation of the future and should get as many eyes as possible to become good quality software

obscuren:
@romanman this seems like a very decent PR, why would you reject this? @ligi is right about the less error proneness.

romanman:
@obscuren Hey Jeff, it's unnecessary generalization
 for gas pricing , in that case I prefer clear simplicity.

ligi:
IMHO it is no generalization - just a helping method that is introduced ( protected ) to be used when the very same pattern is used ( so far 3 times ) - there are more deletions than additions - so it should be simpler ;-)
Also now if the pattern changes you just have to change it in one place not in 3. And it is simper to read this way - just compare:
-            if (data == null) return 1;
-            return 1 + (data.length + 31) / 32 * 1;
+           return getGasForDataAndBase(data,1);

Bitcoinzie:
I agree, it looks like the code is just simplified, as everything is handled once here: 

```
protected long getGasForDataAndBase(byte[] data,int base) {
            if (data == null) return base;
            return base + (data.length + 31) / 32 * base;
}
```

Instead of in each individual method. Subsequent calls are now made to "getGasForDataAndBase" along with an int value for the base and byte [] data. It's the exact same result but with fewer lines making the entire thing easier to understand. So keeping best practices in mind, @ligi's PR looks good to me. Just my 2 cent.

PR #197, opened 2015-01-10
speed improvements, simplification and add test-cases

ligi:


romanman:
@ligi awesome I will go over it first thing tomorrow.

Bitcoinzie:
@ligi Just about imports.. I know it's nitpicky, but organization by 1st ethereum packages(alphabetically), then other libraries(alphabetically), and then java"xx" packages(alphabetically).
Helps keep us all doing things the same way, makes it easier we all know where to expect things to be. And really it's not to difficult to just organize them as you import. BTW thanks for all your work! I enjoy your style.

ligi:
@romanman thanks!
@Bitcoinzie usually I leave this to IntelliJ Idea - do we have importable settings for this - noone should reorder imports by hand ;-)

Bitcoinzie:
@ligi I use netbeans, I have it set up to organize as I go. I believe @cbeams has an IDEA configuration file in the master that organizes import for you. (https://github.com/ethereum/ethereumj/tree/master/.idea)[https://github.com/ethereum/ethereumj/tree/master/.idea] yeah that by hand crap is for the birds ;)

ligi:
@Bitcoinzie ah thanks - I just imported the core and not the whole thing ( not really interested in studio ) - so this file was not picked up as it is in the root. Is now included so all imports should be in right order now - also just updated the PR fixing this

romanman:
https://travis-ci.org/ethereum/ethereumj/builds/46576171

ligi:
sorry - fixed with  59a5baf

ethereumj-core/src/main/java/org/ethereum/vm/DataWord.java
```diff
@@ -32,30 +32,31 @@ public DataWord() {
     }
 
     public DataWord(int num) {
```
romanman: @ligi : that is wrong , the DataWord by 
definition is 32 bytes even if all are zeroes 
fix it in all other places as well

PR #203, opened 2015-01-17
remove typos / add missing modifiers / code cleanup

ligi:


ethereumj-core/src/main/java/org/ethereum/core/Block.java
```diff
@@ -37,11 +37,9 @@
 
     private static final Logger logger = LoggerFactory.getLogger("block");
 
-    public static BigInteger BLOCK_REWARD = BigInteger.valueOf(1500000000000000000L);
-    public static BigInteger UNCLE_REWARD = BLOCK_REWARD.multiply(
```
romanman: I do the breaks for long lines as a standard no need to fix it.

ethereumj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -91,8 +91,8 @@ public void execute() {
 
 
         // FIND OUT THE TRANSACTION TYPE
-        byte[] receiverAddress = null;
-        byte[] code = EMPTY_BYTE_ARRAY;
```
romanman: code should be EMPTY_BYTE_ARRAY

ethereumj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
```diff
@@ -101,14 +101,11 @@ public PeerInfo findOnlinePeer(Set<PeerInfo> excludePeers) {
         worldManager.startPeerDiscovery();
 
         final Set<PeerInfo> peers = worldManager.getPeerDiscovery().getPeers();
-        synchronized (peers) {
```
romanman: Why do you change it ?

ethereumj-core/src/main/java/org/ethereum/net/eth/EthHandler.java
```diff
@@ -212,11 +212,11 @@ public void processStatus(StatusMessage msg, ChannelHandlerContext ctx) throws I
         }
 
         if (!Arrays.equals(msg.getGenesisHash(), Blockchain.GENESIS_HASH)
-                || msg.getProtocolVersion() != EthHandler.VERSION) {
```
romanman: in that particular place static import 
makes it less readable

ethereumj-core/src/main/java/org/ethereum/net/eth/GetBlocksMessage.java
```diff
@@ -72,9 +72,7 @@ public EthMessageCodes getCommand() {
     }
 
     public String toString() {
-        if (!parsed) parse();
```
romanman: why have you removed that line ?

ethereumj-core/src/main/java/org/ethereum/vm/Program.java
```diff
@@ -363,7 +363,7 @@ public void createContract(DataWord value, DataWord memStart, DataWord memSize)
 
         ProgramResult result = null;
 
-        if (programCode != null && programCode.length != 0) {
+        if (programCode.length != 0) {
```
romanman: important null check

romanman:
@ligi : Left some notes, fix it and I will approve the rest

ethereumj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
ligi: because it made no sense before - was syncronized to a local variable - so just removed a nop

ethereumj-core/src/main/java/org/ethereum/net/eth/GetBlocksMessage.java
ligi: getBlockHashes() is doing this already

ethereumj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
ligi: no - it gets overwritten shortly afterwards in any case - assigning  EMPTY_BYTE_ARRAY would have no effect at all.

ethereumj-core/src/main/java/org/ethereum/net/eth/EthHandler.java
ligi: there is no static import - we are inside EthHandler here

ethereumj-core/src/main/java/org/ethereum/vm/Program.java
ligi: it can never be null at this place

ethereumj-core/src/main/java/org/ethereum/core/Block.java
ligi: OK removed it - just thought it is better readable this way

PR #212, opened 2015-01-22
Simplify RLP

ligi:
39 additions and 82 deletions.
this file needs it - approaches 1000 lines ;-)


ligi:
@coveralls either has a bug or is miss-configured when clicking to the details you see a coverage increase for the only changed file as you should expect with this PR - but here it states -0.09%

PR #225, opened 2015-02-23
Minor fixes: make public static field CONFIG final and unquote variable expression

swaldman:
Tiny tiny things...

It's pretty clear that `DEFAULT_TX_APPROVE_TIMEOUT` ought not be quoted.

Letting CONFIG be a nonfinal variable seems a bit loose; any code anywhere could swap out the entire config.


swaldman:
Oh shoot.

I only intended for the first commit to be part of this pull request, but it looks like github has added the work I've put on top of that. The other commits are harmless, but unmotivated at this point. Sorry about that!

I mean to write an alternative implementation of config/SystemProperties.java based on [typesafe-config](https://github.com/typesafehub/config), a pretty great config library that can be compatible with the existing property files, but allows for very expressive configuration formats, easily supports System.properties overrides, and yields an immutable structure which can be exposed publicly without fear of modification. (It's that last bit I really want; I want to experiment with the codebase and have the capacity to add my own config without having to write new accessors for everything.)

Anyway, the only thing that's really burning to be fixed is the mistaken quoting of `DEFAULT_TX_APPROVE_TIMEOUT`. If all this looks annoying, you might just fix that directly.

The pull request fails Travis CI in the same way that my initial clone of the repository fails:

```
test.ethereum.jsontestsuite.GitHubStateTest > stTransactionTest FAILED
    java.lang.AssertionError at GitHubStateTest.java:111

648 tests completed, 1 failed, 47 skipped
```

swaldman:
Closing. I'll redo a very minor pull request for the changes I actually intended to suggest, putting the experimental work in a separate branch. Sorry about all the mess above.

romanman:
@swaldman , don't worry about the travis failures it the tck testing that is failing

PR #227, opened 2015-02-26
Quadratic memory cost

faizkhan00:


ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -210,7 +210,7 @@ else if (oldValue != null && newValue.isZero()) {
             long memoryUsage = (newMemSize.longValue() + 31) / 32 * 32;
             if (memoryUsage > oldMemSize) {
                 memWords = (memoryUsage - oldMemSize) / 32;
-                long memGas = GasCost.MEMORY * memWords;
+                long memGas = GasCost.MEMORY * (memWords + memWords * memWords / 1024);
```
romanman: What test that fix ? can you include back any ?

PR #230, opened 2015-03-04
Add check for halting execution where debit amount is zero and transacti...

faizkhan00:
...on has code to execute


faizkhan00:
Not relevant

PR #240, opened 2015-03-14
Fixed NPE on reset to empty root via trie.setRoot( EMPTY_TRIE_HASH )

swaldman:
On TrieImpl, resetting root by calling `trie.setRoot( EMPTY_TRIE_HASH )` seems like an operation that ought to be okay. But `EMPTY_TRIE_HASH` is interpreted as an ordinary hash value, such that updates of the trie following the reset provoke a NPE as no node with that hash is found. fwiw, here's a test that currently fails and quick fix of the problem.


romanman:
hey @swaldman : ping me on skype : roman.mandeleil.

Thanks.

PR #242, opened 2015-03-16
VMTests/StateTests update + Tier compatibility

faizkhan00:


faizkhan00:
Something I'm confused about here is that Travis has trouble downloading some of testcases, which cases the build to fail, but I don't see it on my local test, so I'm still looking for the root cause of those Travis falls

PR #250, opened 2015-06-03
POC 9 Ethash23

swaldman:
Implemented Ethash23 DAG file generation and mining (The algorithm itself is implemented in Scala, in [consuela](https://github.com/swaldman/consuela), a new dependency. This pull request integrates that implementation with ethereumj.)

Highlights:

* CLIInterface supports a new command line option, -dag, which accepts a blockNumber and generates a DAG file. the DAG files diff as identical to those created by the cpp-ethereum. If you give the JVM heap space sufficient to build the DAG file in-memory (guesstimated as 2.5 times the full size of the DAG), the DAG file will be build concurrently with performance that on my machine rivals or exceeds cpp-ethereum. (Use JVM settings -Xms and -Xmx) Without so much memory, DAGs will stream single-threaded into files, which is slower but very memory efficient. DAG files are cached in accordance with the [Ethash DAG](https://github.com/ethereum/wiki/wiki/Ethash-DAG) spec.

* Per the Ethash 23 spec, if a node is configured to be a miner, it will load a DAG in-memory and mine against that. If a DAG file is cached, it will load that. If no DAG file is cached, it will generate one and attempt to cache it. With each block a node mines or verifies, if the DAG for the next epoch is not already cached, the node will randomly consider beginning to generate and cache that next DAG. Stochastic forward caching mean, when there are many ethereumj nodes, there won't be any synchronized loss of performance due to simultaneous generation of DAGs. Of course, users may avoid any slowing by precaching DAGs in a separate process. In any case, single-threaded streaming of a new DAG takes on the order of 10 mins, a trivial fraction of an epoch, and mining is not paused, only slowed by the loss of a core during that time. The performance impact of stochastic precaching should be negligible.

* A new, highly concurrent `org.ethereum.mine.Miner` is implemented. It is (lightly) tested and mines successfully in `test.ethereum.mine.MinerTest`. In the test library, an adapter is implemented to the previous Miner API so that older tests compile. But those tests remain annotated `@Ignore`, and I'm not quite sure how best to migrate them forward.

I'm sorry to be very out of the loop. I hope that this isn't entirely useless!


romanman:
Hey @swaldman, It is a great surprise, in general I didn't want to put resources on minning for ethereumj but if you did so great job we can include it somehow

swaldman:
(this does include code to verify mixHash and nonce as a light client, if the config param miner is set to `false`. so there's something here for nonminers too, i hope! ;)

swaldman:
(any feedback on why this was closed? no criticism -- this is a gift economy, no strings attached! but i'm curious if there's a reason you decided now to close this.)

romanman:
Hey @swaldman, it was submited on old branch that I removed eventually so the PR was auto closed. No wories.

swaldman:
ok, thanks for the explanation!

let me know if you'd like a PR to a different branch.

romanman:
@swaldman : sure , develop is the running branch of poc-9 
the status is now that we have conflict on 505k block and 
all the focus is now to catch up with the live net which is over 620k
now.

swaldman:
505k is a lot of good blocks! what does the divergence look like? i wish i knew how to get things running enough to help make sense of the difference.

romanman:
@swaldman : not much,  some age cases in HReturn and memory management 
already found good way to deal with that all, how is scala can you 
run all the block chain ?

swaldman:
no, nowhere near that point yet. i don’t have a VM or networking implemented.

romanman:
I see, ok I see,  thanks a lot for your work 
I wish we could arrange a better way to collaborate.

swaldman:
maybe we'll find a way.

i'm excited to hear that ethereumj is working so well against the current blockchain, though. that's huge. congratulations on the great work!

romanman:
@swaldman : thanks

PR #272, opened 2015-07-28
fix GenesisLoader when genesis block is packaged in jar

fjl:
Loading of the genesis block did not work for me when running from a jar file.
I have no idea whether this is a good fix or not, but it resolved the issue for me.


romanman:
:+1:  thanks just also fixed that

PR #328, opened 2015-11-29
 Cleaned unused dependencies.

adrian-tiberius:
### Review of changes:

**Account** - replace WorldManager dependency with Repository
**Wallet** - remove unused WorldManager dependency
**MessageQueue** - replaced WorldManager dependency with EtherumListener
**PeerClient** - replaced WorldManager dependency with EthereumListener
**EthHandler** - replaced WorldManager dependency with EthereumListener and Wallet
**P2PHandler** - replaced WorldManager dependency with EthereumListener and PeerDiscovery
**DiscoveryChannel** - replaced WorldManager dependency with EtherumListener
**MessageCoded** - replaced WorldManager dependency with EthereumListener
**NodeManager** - replaced WorldManager dependency with EthereumListener
**UDPListener** - removed unused WorldManager dependency
**ChannelManager** - removed unused WorldManager, NodeManager, Ethereum dependencies
**EthereumChannelInitializer** - removed unused WorldManager dependency
**PeerServer** - removed unused ChannelManager dependency and replaced WorldManager dependency with EthereumListener
**ShhHandler** - replaced WorldManager dependency with EthereumListener
**BzzHandler**  - replaced WorldManager dependency with EthereumListener
**PeersPool** - replaced Ethereum dependency with ApplicationCotext for PeerClient use

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/sync/PeersPool.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sync/PeersPool.java
@@ -1,7 +1,7 @@
 package org.ethereum.sync;
 
 import org.ethereum.db.ByteArrayWrapper;
-import org.ethereum.facade.Ethereum;
+import org.ethereum.net.client.PeerClient;
 import org.ethereum.net.eth.EthVersion;
 import org.ethereum.net.rlpx.Node;
 import org.ethereum.net.server.Channel;
@@ -10,6 +10,7 @@ import org.ethereum.util.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
 import org.springframework.stereotype.Component;
 
 import javax.annotation.Nullable;
@@ -55,7 +56,7 @@ public class PeersPool implements Iterable<Channel> {
     private final Map<String, Long> pendingConnections = new HashMap<>();
 
     @Autowired
-    private Ethereum ethereum;
+    private ApplicationContext ctx;
 
     @PostConstruct
     public void init() {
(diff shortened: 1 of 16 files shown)


ethereumj-core/src/main/java/org/ethereum/sync/PeersPool.java
```diff
@@ -193,7 +194,13 @@ public void connect(Node node) {
         }
 
         synchronized (pendingConnections) {
-            ethereum.connect(node);
```
Nashatyrev: Hi Adrian, 
why do we need this change, that looks like a code duplication?

adrian-tiberius: Hi Nashatyrev,

I tried to solve a dependency cycle ( Ethereum -> WorldManager -> SyncManager -> PeersPool -> Ethereum ). This was the most non-intrusive way I thought of doing that, at the cost of small code duplication. 

To avoid code duplication entirely, I thought about having the PeersPool handle peer client connections (to me PeersPool should have all the functionality for connecting to a peer inside it instead of calling the top class, otherwise I don't see the point of having a connect method on it) and have PeersPool as a dependency of either EthereumImpl or WorldManager.

My understanding of how all these blocks interact and work with each other is still somewhat limited, so I'm trying to make as few changes as possible in each commit, so that it is easier to review and spot if my understandings of things take me on a wrong road.

Nashatyrev: Hi Adrian, 
sounds reasonable. 
Could we move the connect method from the `EthereumImpl` to the `ChannelManager` and use it from both places? It seems to me the most appropriate place?

adrian-tiberius: Hi Nashatyrev,
I reverted the changes of PeersPool for now and will address this issue in a new pull request.
After looking further into this, my proposal is to create a new class PeerClientManager which would be referenced by any other component which requires peer client connections. I will add a pull request shortly to reflect these changes so you can give me your oppinion on the matter.

Nashatyrev: Cool, I think this would be a suitable option!
Thanks for contributing!

PR #332, opened 2015-12-02
Added PeerClientManager to handle PeerClient connections.

adrian-tiberius:
### PeerClient connection changes implemented in this pull request
What I'm proposing is a new class **PeerClientManager** which is referenced by any other component in need of a **PeerClient** connect ( including the active peer client logic ). I also added some event listeners to **PeerClient** and **EthereumChannelInitializer** to signal to the **PeerClientManager** when a channel is initialized/closed. The **PeerClientManager** will, in turn, signal the **ChannelManager** of these events (instead of having the **ChannelManager** dependency in **EthereumChannelInitializer**).
**PeerConnectionTester**, **Hive**, **EthereumImpl** and **PeersPool** are the components that use the new manager.

### Alternative ideea not implemented in this pull request
As an alternative we could have **PeerClient** as a singleton component, instead of a prototype, to replace this manager class or remove **PeerClient** class and have its logic moved inside the manager class. I'm sure I'm missing something about **PeerClient** being a prototype, but I don't see any value for it. It doesn't have any state that I can see which could be used, and its not kept anywhere as an instance once its created (except for activePeer component which is something I don't understand since, from what I can see, using connect on it its the same as creating a new peer and using connect on that). But like I said I think I'm missing something about how this is used as a prototype, so any hint to its purpose would be greatly apreciated.


Nashatyrev:
Hi Adrian, 
PeerClientManager is the good shot. 
However I would avoid complicating the execution flow with propagating events forth and back via a number of listeners. The main focus is `Channel` no matter how it was created (via PeerClient or PeerServer, they are just 'lightweight' connectors) and `ChannelManager`. 
Is the ChannelManager dependency in EthereumChannelInitializer so bad? 
I would add the only listener one might be interested in to the ChannelManager (see my update: https://github.com/ethereum/ethereumj/commit/c926f20473f2473be3d31b79773f020b75978411)
Cheers.

adrian-tiberius:
Hi Nashatyrev,
Thank you for the feedback. You are right, it is cleaner with your update. 
I removed ChannelManager dependency from PeerClientManager as it is no longer needed there.
I also removed the SyncManager dependency from ChannelManager by using the new ChannelListener functionality(fd0fd02).

Nashatyrev:
Hi @adrian-tiberius,
sorry for leaving your PR without any attention for so long time... 
We are currently making some more refactorings and stabilizing the core to make a release. I'll be back to you after all these changes are made

Issue #360, opened 2016-03-28
Failing tests in “develop” branch


loxal:
These tests are failing when I run the tests locally on my Mac:

> org.ethereum.core.TransactionTest#multiSuicideTest
> org.ethereum.solidity.CompilerTest#simpleTest

The code cannot deal with an empty string that is returned but I guess the problem is that an empty string is returned at all. Also org.ethereum.solidity.CompilerTest#simpleTest should get an assertion. I’ve fixed the tests in a way that they way because of not fulfilled assertions but not because of some unexpected behavior. A corresponding PR follow in minutes :)

Nashatyrev:
Thanks for PR! 
Solc on Mac should work from 1.2.0

PR #361, opened 2016-03-28
#360

loxal:


ethereumj-core/src/test/java/org/ethereum/core/TransactionTest.java
```diff
@@ -553,19 +553,23 @@ public void multiSuicideTest() throws IOException, InterruptedException {
...
-        ProgramResult programResult = executeTransaction(blockchain, tx1);
+            Transaction tx1 = createTx(blockchain, sender, contractAddress, callData);
+            ProgramResult programResult = executeTransaction(blockchain, tx1);
 
-        // suicide of a single account should be counted only once
-        Assert.assertEquals(programResult.getFutureRefund(), 24000);
+            // suicide of a single account should be counted only once
+            Assert.assertEquals(programResult.getFutureRefund(), 24000);
+        } else {
+//            Assert.fail();
```
Nashatyrev: This needs to be uncommented probably?

ethereumj-core/src/test/java/org/ethereum/solidity/CompilerTest.java
```diff
@@ -16,17 +16,20 @@
...
-                "  int i1;" +
-                "  function i() returns (int) {" +
-                "    return i1;" +
-                "  }" +
-                "}";
+                        "  int i1;" +
+                        "  function i() returns (int) {" +
+                        "    return i1;" +
+                        "  }" +
+                        "}";
```
Nashatyrev: Those indents (likely inserted by IDE) make the contract source less readable

ethereumj-core/src/test/java/org/ethereum/core/PendingStateTest.java
```diff
@@ -119,7 +119,7 @@ public void changeForkTest() {
         pendingState.processBest(b1);
         assertEquals(pendingState.getPendingTransactions().size(), 0);
 
-        Block b1_ = blockchain.createNewBlock(blockchain.getBestBlock(), Arrays.<Transaction>asList(), Collections.EMPTY_LIST);
+        Block b1_ = blockchain.createNewBlock(blockchain.getBestBlock(), Collections.EMPTY_LIST, Collections.EMPTY_LIST);
```
Nashatyrev: The `Collections.emptyList()` is more progressive I believe

ethereumj-core/src/test/java/org/ethereum/core/TransactionTest.java
loxal: Yes, you’re right. Done.

ethereumj-core/src/test/java/org/ethereum/solidity/CompilerTest.java
loxal: I made good experience having and auto-format hook before commit. This assures a uniformity of source code format. You can write messy formatted code and it will be committed in a format that many developers agree is a readable format. But I can revert that although I do not think it it’s worse. Adjusted.

ethereumj-core/src/test/java/org/ethereum/core/PendingStateTest.java
loxal: Actually I’ve tried to do it all over the code but I get a compile time error. There is a significant number of unchecked casts all over the code (just compile with `-Xlint:unchecked`). If we fix them, we might be able to use this static method. In this case I’ve reverted the the `emptyList()` that I’ve introduced in a previous commit.

PR #363, opened 2016-03-30
Generalizations to ECKey

dwhjames:
The primary motivation here is to generalize the
cryptography routines in ECKey so that the
Java Cryptography Architecture with pluggable
providers can be used.
This is especially relevant for situations where
the private key material is not resident in
JVM process memory space, and instead held in
a hardware security module (HSM).

To enable this, the private member
```java
private BigInteger priv;
```
has been replaced with an abstract private key
to encapsulate the private key material.
```java
private final PrivateKey privKey;
```

One additional private member has been added
```java
private final Provider provider;
```
which pins the provider to use when computing
ECDSA signature with the private key.

All other changes to ECKey here follow from these
two changes.

I believe the public interface of ECKey has been
preserved here, although I do have followup
suggestions for changes to the public interface.

And existing code-paths will effectively remain
identical. The BouncyCastle (well, SpongyCastle)
cryptography will still be used if the
user does not elect to override it with a
different security provider.

The line and branch test coverage has been increased
in ECKey as well.


dwhjames:
Please let me know if you’d like additional explanation on the motivation for this PR and my intended use cases!

ethereumj-core/src/main/java/org/ethereum/crypto/ECKey.java
```diff
@@ -445,6 +540,31 @@ public static boolean validateComponents(BigInteger r, BigInteger s, byte v) {
             return true;
         }
 
+        public static ECDSASignature decodeFromDER(byte[] bytes) {
```
dwhjames: The header javadoc for this class already says
```java
 * This code is borrowed from the bitcoinj project and altered to fit Ethereum.<br>
 * See <a href="https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/com/google/bitcoin/core/ECKey.java">
 * bitcoinj on GitHub</a>.
```

but I just realized that I wasn’t explicit here…
this decoding for a DER encoded ECDSA signature was also copied verbatim from BitcoinJ’s `ECKey`

romanman:
@dwhjames : thank you, let us to review
that and get back to you

dwhjames:
> @dwhjames : thank you, let us to review
> that and get back to you

Thanks for the quick comment.

Once you’ve had a chance to review, I have several followups I’d like to discuss.

dwhjames:
Woah, looks like I missed something in local testing :) investigating now

romanman:
cool take your time

dwhjames:
Ok, here’s the explanation for the test failure.
I humbly submit that this was a bug in the test ;)

The failing test is
```java
    @Test  // cpp keys demystified
    public void test13() throws Throwable {

//        us.secret() a4627abc2a3c25315bff732cb22bc128f203912dd2a840f31e66efb27a47d2b1
//        us.public() caa3d5086b31529bb00207eabf244a0a6c54d807d2ac0ec1f3b1bdde0dbf8130c115b1eaf62ce0f8062bcf70c0fefbc97cec79e7faffcc844a149a17fcd7bada
//        us.address() 47d8cb63a7965d98b547b9f0333a654b60ffa190


        ECKey key = ECKey.fromPrivate(Hex.decode("a4627abc2a3c25315bff732cb22bc128f203912dd2a840f31e66efb27a47d2b1"));

        String address = Hex.toHexString(key.getAddress());
        String pubkey  = Hex.toHexString(key.getPubKeyPoint().getXCoord().getEncoded()) +  // X cord
                         Hex.toHexString(key.getPubKeyPoint().getYCoord().getEncoded());   // Y cord

        log.info("address: " + address);
        log.info("pubkey: " + pubkey);

        assertEquals("47d8cb63a7965d98b547b9f0333a654b60ffa190", address);
        assertEquals("caa3d5086b31529bb00207eabf244a0a6c54d807d2ac0ec1f3b1bdde0dbf8130c115b1eaf62ce0f8062bcf70c0fefbc97cec79e7faffcc844a149a17fcd7bada", pubkey);
    }
```

The bug is in
```java
key.getPubKeyPoint().getXCoord().getEncoded()
```
and similarly for the `getYCoord()`

The previous implementation of `ECKey#fromPrivate` was
```java
return new ECKey(privKey, compressPoint(CURVE.getG().multiply(privKey)));
```
The important issue being that it compressed the point (the point compression and decompression in the codebase, and why I think this is erroneously used is a topic for another time).

To keep the elliptic curve operations efficient, internally, the ECPoint has Z-coords as well as X and Y.
There is a warning in the javadoc for `getXCoord` and `getYCoord`
https://github.com/rtyley/spongycastle/blob/6bf34b6a3b133081c19b664a7a8f2b4a3c2a854f/core/src/main/java/org/spongycastle/math/ec/ECPoint.java#L147-L150
To get back to the affine coordinate space, to get the X and Y values one expects, the point needs to be normalized.
`getX` and `getY` are now deprecated, so a `normalize` followed by `getAffineXCoord` and `getAffineYCoord` are needed to safely get the expected coordinates in the expected coordindate space.

The reason this use of `getXCoord` and `getYCoord` escaped notice before was that `compressPoint` had the additional effect of forcing a normalization.
Bouncy Castle is doing `k * G` as efficiently as possible, but then `compressPoint` is doing a round-trip through serialization and de-serialization, thus doing additional expensive elliptic curve operations and ending up with a ECPoint already in normalized form.

My fix will be to correct the test to
```java
        ECPoint pubPoint = key.getPubKeyPoint().normalize();
        String pubkey  = Hex.toHexString(pubPoint.getAffineXCoord().getEncoded()) +  // X cord
                         Hex.toHexString(pubPoint.getAffineYCoord().getEncoded());   // Y cord
```
to remove the use of the unsafe methods on ECPoint

romanman:
sounds good , we probably are using in more 
places that not normalized cords, but it need to 
be tested also by running the live blockchain to see
that it has full support

romanman:
@dwhjames : what is your email , will invite you to a better 
chat room, or autoinvite on www.ether.camp

dwhjames:
> sounds good , we probably are using in more 
> places that not normalized cords, but it need to 
> be tested also by running the live blockchain to see
> that it has full support

Directly accessing the coords in an ECPoint is rather unusual, and I would be surprised if you did this outside of test code.
BouncyCastle forces a normalization step at the point that you call `getEncoded`
https://github.com/rtyley/spongycastle/blob/6bf34b6a3b133081c19b664a7a8f2b4a3c2a854f/core/src/main/java/org/spongycastle/math/ec/ECPoint.java#L462

So maybe the test would be better if it were refactored to get the encoded form, rather than attempting to do the encoding itself.
The reason the assertion on the address succeeded is because internally `ECKey` is calling `getEncoded` on the public point to get it in uncompressed form.

dwhjames:
I’ve just tacked on another commit, dd053ac, to clean up the handling of `ECKey` throughout the codebase. As this PR has been hanging out for a while, I thought I’d get it all done in one PR.

dwhjames:
And one more commit I had saved up: 666225b

ethereumj-core/src/main/java/org/ethereum/crypto/ECKey.java
```diff
@@ -597,14 +885,18 @@ public static ECKey signatureToKey(byte[] messageHash, String signatureBase64) t
      */
     public byte[] decryptAES(byte[] cipher){
 
-        if (priv == null)
+        if (privKey == null) {
             throw new MissingPrivateKeyException();
+        }
+        if (!(privKey instanceof BCECPrivateKey)) {
+            throw new UnsupportedOperationException("Cannot use the private key as an AES key");
+        }
```
dwhjames: This method `ECKey#decryptAES(byte[])` should really just be deleted. The only place this method is used is in a standalone test in `ECKeyTest`.
It is rather odd to take the scalar value of a EC private key and use it as a AES key.
I couldn’t deduce from the git history why this method was added in the first place.

ethereumj-core/src/main/java/org/ethereum/crypto/ECKey.java
```diff
@@ -345,9 +516,13 @@ public ECPoint getPubKeyPoint() {
      * @throws java.lang.IllegalStateException if the private key bytes are not available.
      */
     public BigInteger getPrivKey() {
-        if (priv == null)
+        if (privKey == null) {
+            throw new MissingPrivateKeyException();
+        } else if (privKey instanceof BCECPrivateKey) {
+            return ((BCECPrivateKey) privKey).getS();
```
Nashatyrev: I see that `BCECPrivateKey.getS()` and `BCECPrivateKey.getD()` return the same value. Here both of them are used, is it intentionally? 
I'm also wondering if this is the same matter which just has different names in different interfaces or there is a subtlety?

dwhjames: They are the same.
The Java security interface uses the name `S` for the private key value:
https://docs.oracle.com/javase/8/docs/api/java/security/interfaces/ECPrivateKey.html#getS--
The Bouncy Castle interface uses the name `D` for the private key value:
http://www.bouncycastle.org/docs/docs1.5on/org/bouncycastle/jce/interfaces/ECPrivateKey.html#getD()

But ultimately it’s the same value:
https://github.com/bcgit/bc-java/blob/3d610624e9e87b8abee4998dd899259d8511449c/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/ec/BCECPrivateKey.java#L339-L347


The analogue occurs for the public point:
https://docs.oracle.com/javase/8/docs/api/java/security/interfaces/ECPublicKey.html#getW--
http://www.bouncycastle.org/docs/docs1.5on/org/bouncycastle/jce/interfaces/ECPublicKey.html#getQ()
https://github.com/bcgit/bc-java/blob/3d610624e9e87b8abee4998dd899259d8511449c/prov/src/main/java/org/bouncycastle/jcajce/provider/asymmetric/ec/BCECPublicKey.java#L330-L343
although there the `ECPoint`s are different nominal types, despite again being the same pair of values

dwhjames: Oh, I see your original point through…
I used a mix of `.getS()` and `.getD()` in various methods.
No, that wasn’t a conscious choice. I can certainly clean that up to be consistent.

dwhjames: fixed in 1b1be96

ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
```diff
@@ -253,9 +249,11 @@ public ECKey getKey() {
     }
 
     public void sign(byte[] privKeyBytes) throws MissingPrivateKeyException {
-        byte[] hash = this.getRawHash();
-        ECKey key = ECKey.fromPrivate(privKeyBytes).decompress();
-        this.signature = key.sign(hash);
+        sign(ECKey.fromPrivate(privKeyBytes));
```
dwhjames: note:
I think `Transaction#sign(byte[])` should be deprecated and removed in the long-run.

Nashatyrev:
Something was broken with the commit https://github.com/ethereum/ethereumj/pull/363/commits/dd053ac185f0ec72d43348d8a534fb9139ef975c
`samples.TestNetSample` can't AuthInitiate to Go peers 
`samples.PrivateNetSample` one Java peer couldn't connect to another one (AuthInitiate can't be decrypted)
Trying to find out the reason

dwhjames:
@Nashatyrev could you describe how to reproduce your test?
I’m looking too.

dwhjames:
Do you a suspect a break in `EncryptionHandshake` for RLPx?

Nashatyrev:
@dwhjames You may just run one of those samples and see the log for handshake messages. 
`TestNetSample` just got disconnect from remote peers. 
`PrivateNetSample` runs two Java peers and one tries to connect to another but the latter can't decrypt AuthHandshake 
Just wondering why some peers on the Homestead network works fine, i.e. we may connect to them..

dwhjames:
@Nashatyrev, I managed to replicate the issue by running those samples.

I can’t spot anything obviously wrong, so I’m going to expand the test suite for things like `EncryptionHandshakeTest` and see if I can replicate the bug in a simulation.

Shout if you find the root cause.

Nashatyrev:
Hey Daniel! 
Was at Consesus events last week, so had no time to get back to this PR. 
Any news from your side so far?

dwhjames:
Hey @Nashatyrev, sorry, I haven’t had much time recently to resurrect this.
I’m still planning to expand the tests around peer handshakes. I’m guessing the issues lies somewhere here.
I was surprised I managed to break something as Ethereum shouldn’t need any EC point compression (in comparison to Bitcoin). Maybe something subtle was disturbed by my changes.

ethereumj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
```diff
@@ -98,11 +98,7 @@ public void initiate(ChannelHandlerContext ctx) throws Exception {
 
         nodeId = myKey.getNodeId();
 
-        byte[] remotePublicBytes = new byte[remoteId.length + 1];
-        System.arraycopy(remoteId, 0, remotePublicBytes, 1, remoteId.length);
-        remotePublicBytes[0] = 0x04; // uncompressed
-        ECPoint remotePublic = ECKey.fromPublicOnly(remotePublicBytes).getPubKeyPoint();
-        handshake = new EncryptionHandshake(remotePublic);
+        handshake = new EncryptionHandshake(ECKey.fromNodeId(nodeId).getPubKeyPoint());
```
Nashatyrev: Should be `remoteId` here

Nashatyrev:
With the latest fix all seems to work fine

ethereumj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
dwhjames: Ah! Good catch!

dwhjames:
> With the latest fix all seems to work fine
> Can you please make a commit?

Yup, I’ll commit that shortly.

dwhjames:
Seems like there should be an easy test to exercise this code path. Do you have any suggestions for where/what to add?

ethereumj-core/src/main/java/org/ethereum/crypto/ECKey.java
dwhjames: @Nashatyrev, are you open to removing `ECKey#decryptAES(byte[])`?

Nashatyrev: Let's just mark it deprecated, not sure if there no projects using it for some reason

Nashatyrev:
> Seems like there should be an easy test to exercise this code path. Do you have any suggestions for where/what to add?

Network related stuff is a bit complicated to cover with short running JUnit tests. We are running some long term scenarios separately for that.

ethereumj-core/src/main/java/org/ethereum/crypto/ECKey.java
dwhjames: done in 4f80b8d

ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
dwhjames: marked in 4f80b8d

dwhjames:
> Network related stuff is a bit complicated to cover with short running JUnit tests. We are running some long running scenarios separately for that.

Ok. Feel free to ping me if you would like contributions in the future to expand the test coverage, once you have a plan of action.

Nashatyrev:
Daniel, can we do some small sample for this generalization PR which demonstrates the use case?

dwhjames:
> Daniel, can we do some small sample for this generalization PR which demonstrates the use case?

@Nashatyrev, good question.

It might be a little tough to demo a compelling use of the Java Security Provider mechanism in an uncontrived way.

Swapping from the BouncyCastle elliptic curve cryptography back-end to the Sun backend
https://docs.oracle.com/javase/7/docs/technotes/guides/security/SunProviders.html#SunJCEProvider
is not particularly interesting as it’s still just crypto routines within the JVM process. (Technically the BouncyCastle cryptography is in pure Java, whereas the core of the Sun cryptography is in native code in the JRE.)


One can simulate a hardware security module with a project like:
https://github.com/opendnssec/SoftHSMv2
And then using the Sun PKCS#11 provider to interface with it.
https://docs.oracle.com/javase/7/docs/technotes/guides/security/SunProviders.html#SunPKCS11Provider
So the cryptography would now be out of the JVM process, but still in software.
SoftHSMv2 is a great project—very useful for testing purposes—but setting it up for an easily runnable EthereumJ demo might be rather tricky.

The true use case of the generalizations to `ECKey` is to be able to hook EthereumJ up to one of these:

- http://www.safenet-inc.com/data-encryption/hardware-security-modules-hsms/luna-hsms-key-management/
- https://www.thales-esecurity.com/products-and-services/products-and-services/hardware-security-modules
- https://www.ultra-aep.com/ultra-safe-overview

:)

Nashatyrev:
Daniel, thanks for valuable contribution!

dwhjames:
> Daniel, thanks for valuable contribution!

Happy to contribute.

PR #374, opened 2016-04-07
Code cleanup 

loxal:
In order to reduce compiler warnings and code smells.

* more consistent usage of append 
* comparing objects w/o ==
* public modifier is superfluous in interface definitions, also no abstract methods there
* fix erroneous overriding (potential runtime issues) of equals (if an obj is an instance of something, it cannot be null)
* fix clone signature
* make pseudo-constants actual constants
* private methods cannot be overridden, hence making them final is just noisy


ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
```diff
@@ -777,7 +790,7 @@ private void addReward(Block block) {
         if (block.getUncleList().size() > 0) {
             for (BlockHeader uncle : block.getUncleList()) {
                 track.addBalance(uncle.getCoinbase(),
-                        new BigDecimal(BLOCK_REWARD).multiply(BigDecimal.valueOf(8 + uncle.getNumber() - block.getNumber()).divide(new BigDecimal(8))).toBigInteger());
+                        new BigDecimal(BLOCK_REWARD).multiply(BigDecimal.valueOf(8 + uncle.getNumber() - block.getNumber()).divide(new BigDecimal(8), RoundingMode.HALF_UP)).toBigInteger());
```
Nashatyrev: Alex, can you please change this to `BigInteger` calculation. There is no place for decimal. 
Thanks we always had even `BLOCK_REWARD`, else we definitely get a conflict here.

loxal:
@Nashatyrev the build fails because of a failing test.
> org.ethereum.jsontestsuite.GitHubBlockTest#runBCRPCAPITest
I guess the test assumes Git commits that are not present in my forked repository.

ethereumj-core/src/main/java/org/ethereum/datasource/redis/RedisSet.java
```diff
@@ -39,6 +38,7 @@ public boolean contains(final Object o) {
         return pooledWithResult(new Function<Jedis, Boolean>() {
             @Override
             public Boolean apply(Jedis jedis) {
+                if(o instanceof Boolean) System.out.println("");
```
Nashatyrev: This is mostly deprecated class but anyways... :)

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
loxal: My fault, apparently the rounding strategy came from an unmerged (and broken) PR. I’ve also changed BigDecimal to BigInteger.

loxal: Assuming the int-based calculation is implemented as desired, we need to adjust the tests.

Nashatyrev: What tests need to be adjusted?

loxal: The CI job fails because of the seitch from BigDecimal to BigInteger.

Nashatyrev:
```
BLOCK_REWARD.multiply(BigInteger.valueOf(8 + uncle.getNumber() - block.getNumber())).divide(BigInteger.valueOf(8))
```

Nashatyrev: If JSON test (TCK in fact) fails then our implementation need to be adjusted most likely :)

loxal:
@Nashatyrev ok, I fixed the broken build incl. the conversion to BigInteger that you have proposed.

Nashatyrev:
Cool, thanks Alex!

PR #376, opened 2016-04-10
show how to actually build & run a self-containing JAR

loxal:
Actually I missed this the most as I tried to run EthereumJ from the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,16 @@ or gradle:
 
 As a starting point for your own project take a look at https://github.com/ether-camp/ethereumj.starter
 
+##### Building an executable JAR
+```
+git clone https://github.com/ethereum/ethereumj
+cd ethereumj
+cp ethereumj-core/src/main/resources/ethereumj.conf ethereumj-core/src/main/resources/user.conf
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,9 +42,11 @@ As a starting point for your own project take a look at https://github.com/ether
...
 ```
-> git clone https://github.com/ethereum/ethereumj
-> cd ethereumj
-> ./gradlew run [-PmainClass=<sample class>]
+git clone https://github.com/ethereum/ethereumj
+cd ethereumj
+vim ethereumj-core/src/main/resources/ethereumj.conf # adjust ethereumj.conf to your needs
+./gradlew clean shadowJar
+java -jar ethereumj-core/build/libs/ethereumj-core-*-all.jar
 ```
```
Nashatyrev: Alex, can you please make this a separate topic:
`Building and running self-containing JAR` 
And please adjust the line regarding `ethereumj.conf` or may be remove it since configuring is described in a separate topic

loxal: I’ve addressed your remarks. As I said, this was the part I missed most in the entire `README`. I guess many other users might look for the same solution. Also I’d suggest to provide those executable JARs with every release as this is the only part users care about.

PR #377, opened 2016-04-10
Update dependencies

loxal:
update major dependencies that have not changed their API to reduce security risks and improve reliability

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,11 @@ As a starting point for your own project take a look at https://github.com/ether
 
 ##### Running from command line:
 ```
-> git clone https://github.com/ethereum/ethereumj
-> cd ethereumj
-> ./gradlew run [-PmainClass=<sample class>]
+git clone https://github.com/ethereum/ethereumj
+cd ethereumj
+vim ethereumj-core/src/main/resources/ethereumj.conf # adjust ethereumj.conf to your needs
+./gradlew clean shadowJar
+java -jar ethereumj-core/build/libs/ethereumj-core-*-all.jar
 ```
 
 ##### Optional samples to try:
--- a/ethereumj-core/build.gradle
+++ b/ethereumj-core/build.gradle
@@ -100,17 +100,17 @@ test {
 }
 
 ext {
-    slf4jVersion = '1.7.7'
+    slf4jVersion = '1.7.21'
     leveldbVersion = '0.7'
     scastleVersion = '1.51.0.0'
     log4jVersion = '1.2.17'
-    springVersion = '4.2.0.RELEASE'
-    hibernateVersion = '4.3.7.Final'
-    junitVersion = '4.11'
+    springVersion = '4.2.5.RELEASE'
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -42,9 +42,11 @@ As a starting point for your own project take a look at https://github.com/ether
 
 ##### Running from command line:
 ```
-> git clone https://github.com/ethereum/ethereumj
-> cd ethereumj
-> ./gradlew run [-PmainClass=<sample class>]
+git clone https://github.com/ethereum/ethereumj
+cd ethereumj
+vim ethereumj-core/src/main/resources/ethereumj.conf # adjust ethereumj.conf to your needs
```
Nashatyrev: `resources/ethereumj.conf` is intended for default config values and is not subject to be changed. 
There is a plenty of ways to supply the user config including `user.conf` as a resource

README.md
```diff
@@ -42,9 +42,11 @@ As a starting point for your own project take a look at https://github.com/ether
 
 ##### Running from command line:
 ```
-> git clone https://github.com/ethereum/ethereumj
-> cd ethereumj
-> ./gradlew run [-PmainClass=<sample class>]
```
Nashatyrev: I would not delete this line - it's an option as well

ethereumj-core/build.gradle
```diff
@@ -139,7 +139,7 @@ dependencies {
     compile "com.h2database:h2:1.4.187"
     compile "com.typesafe:config:1.2.1"
 
-    compile "org.mapdb:mapdb:2.0-beta12"
+    compile "org.mapdb:mapdb:2.0-beta13"
 
     compile "org.slf4j:slf4j-log4j12:${slf4jVersion}"
     compile "log4j:apache-log4j-extras:${log4jVersion}"
```
Nashatyrev: We decided that we would better avoid unreasonable lib versions upgrades due to possible regressions unless there are desired bug fixes or well known security fixes

README.md
loxal: But then I’d put it into another section as otherwise it would confuse sb. who just wants to build a JAR.

ethereumj-core/build.gradle
loxal: Does this apply for all dependencies? Usually e.g. Spring provides security related fixes with some of the minor updates. Also other libraries might provide security related fixes that are not even announced by their vendors. I would feel uncomfortable to run a server that uses outdated libraries. Deserialization issues can compromise the underlying machine.

README.md
loxal: @Nashatyrev I’ve created a PR without the dependency updates for `develop` #376. Same code regarding README.md. 

> ./gradlew run [-PmainClass=<sample class>]
…is not gone as below there are several examples using this pattern.

ethereumj-core/build.gradle
Nashatyrev: Will will probably update the Spring and Netty libs soon but this will be a separate task with extensive testing.

PR #384, opened 2016-04-19
to give a better ECKey test case

mhoudg:
use `new BigInteger(privString, 16) to avoid a negative privateKey
integer

### DIFF
--- a/ethereumj-core/src/test/java/org/ethereum/crypto/ECKeyTest.java
+++ b/ethereumj-core/src/test/java/org/ethereum/crypto/ECKeyTest.java
@@ -33,21 +33,21 @@ import static org.junit.Assert.*;
 public class ECKeyTest {
     private static final Logger log = LoggerFactory.getLogger(ECKeyTest.class);
 
-    private String privString = "3ecb44df2159c26e0f995712d4f39b6f6e499b40749b1cf1246c37f9516cb6a4";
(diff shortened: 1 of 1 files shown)


mhoudg:
I think `new BigInteger(privString, 16)` is a better example than `new BigInteger(Hex.decode(privString)) `. In case that the first character in privString is between 8 to f, 

For example:

    String privString = "3ecb44df2159c26e0f995712d4f39b6f6e499b40749b1cf1246c37f9516cb6a4";
    System.out.println(new BigInteger(privString, 16));
    System.out.println(new BigInteger(Hex.decode(privString)));

prints the same result:

    28402541900644447169288774227207574684351244670225019835538386550149190825636
    28402541900644447169288774227207574684351244670225019835538386550149190825636

while in this case:

    String privString = "c85ef7d79691fe79573b1a7064c19c1a9819ebdbd1faaab1a8ec92344438aaf4";
    System.out.println(new BigInteger(privString, 16));
    System.out.println(new BigInteger(Hex.decode(privString)));

prints different result:

    90630363884335538722706632492458228784305343302099024356772372330524102404852
    -25161725352980656700864352516229679068964641363541539682685211677389027235084

Notice that the Hex.decode(privString) will get an negative number that should not be used.

Nashatyrev:
Thanks!

PR #387, opened 2016-04-21
Remove global static `SystemProperties`, make configurable

stephenjudkins:
NOTE: don't merge this yet. We still have a couple regressions to fix. Our intent is to give you a heads up a large incoming change and begin discussing any concerns.

`SystemProperties` was previously passed through DI in many, though not all, instances. With this change, `SystemProperties` is injected and the global instance has been removed. This gives users the ability to start multiple differently configured instances for unit testing purposes, or to better integrate with another configuration system for their app.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/Start.java
+++ b/ethereumj-core/src/main/java/org/ethereum/Start.java
@@ -1,13 +1,14 @@
 package org.ethereum;
 
 import org.ethereum.cli.CLIInterface;
+import org.ethereum.config.SystemProperties;
 import org.ethereum.facade.Ethereum;
 import org.ethereum.facade.EthereumFactory;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
 
-import static org.ethereum.config.SystemProperties.CONFIG;
+
 
 /**
  * @author Roman Mandeleil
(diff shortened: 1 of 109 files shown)


Nashatyrev:
Hey, Stephen
Thanks for you PR!
Did you see that sample on how different configs can be used in a single VM: https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/samples/PrivateMinerSample.java
We actually would like to avoid propagating config to all the instances and utilize the DI capability.
There are still a number of static `CONFIG` occurrences, but we would like to make them DI'ed as well.

ewiner:
Hey Anton - 

While it's easy to call `EthereumFactory.createEthereum()` multiple times using differently configured `SystemProperties` objects, some of the inner components don't respect the `SystemProperties` you passed in, and instead look at the global `SystemProperties.CONFIG`.  This can cause some very strange behavior as you might imagine, as different bits of code are looking at different config.

For instance, when I tried to configure EthereumJ to talk to the a private blockchain using a custom config passed in to `EthereumFactory.createEthereum()`, it failed to validate blocks because `BlockHeader.calcDifficulty()` was using the global `CONFIG` instead of the one I passed in.  At this point, we're not even trying to get multiple different configs loaded in a single VM - just trying to get one consistent custom config used throughout the EthereumJ code.

I think it makes a lot of sense to work towards using DI to propagate the config everywhere instead of having to pass it around from object to object.  However, I also think this PR is a step in that direction - it's getting the right config into the places it needs to be, so it's more clear what needs to be built into the DI system down the road.  Does that make sense as a plan?

stephenjudkins:
In all the places where DI was previously used to construct an object, DI is now being used to pass in the `SystemProperties` instance. In many instances I replaced the default constructor with one with appropriate arguments for easier testing purposes, but used the `@Autowired` annotation so Spring would continue to be using DI to inject everything.

The only instances where we are now explicitly passing in `SystemProperties` instance is when existing classes were constructed without Spring's help. I wanted to make this PR as minimal as possible, and bringing these classes under DI would require a good deal more work.

I think this is a major step towards the ultimate goal of constructing even more components (when appropriate) with DI.

Nashatyrev:
Hi, guys! 
I've made the alternative change addressing the issue in a bit more conservative way (less config propagation and setting the default config for Autowired fields for standalone unit tests)
If you want you may check this branch: https://github.com/ethereum/ethereumj/tree/feature/no-static-config

Nashatyrev:
Finally merged the branch into develop. 
I've made less invasive fix - the static config is still available to use and is used by all the non-Spring tests. 
Normally the static config is used in 99% of scenarios, so no need to force using 'config' bean only or pass config through all the classes. 
I've also added `SystemProperties.setUseOnlySpringConfig` which forces `null` static config and checked that Spring based scenarios work well, which means they doesn't rely on static config and use per AppContext config only

PR #389, opened 2016-04-22
Fix for `org.ethereum.facade.Ethereum.getPendingState()`

ewiner:
`org.ethereum.facade.Ethereum.getPendingState()` currently fails with a `ClassCastException`, because it returns a `RepositoryTrack` object and tries to cast it into a `org.ethereum.facade.Repository`.

This change makes `RepositoryTrack` extends `facade.Repository` so the cast will succeed.


Nashatyrev:
Thanks!
We still need to revisit all facade interfaces

PR #393, opened 2016-04-25
squid:S1213 - The members of an interface declaration or class should…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S1213 - The members of an interface declaration or class should appear in a pre-defined order

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S1213

Please let me know if you have any questions.

M-Ezzat

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/Block.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Block.java
@@ -34,6 +34,11 @@ public class Block {
 
     private BlockHeader header;
 
+    private StringBuffer toStringBuff = new StringBuffer();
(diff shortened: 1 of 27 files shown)


romanman:
@m-ezzat : Dear Mohamed , thank you for your hard work 
and motivation to have all code correct by the Java style guide. 
However we have feedback from the community and see that 
the style guide is not a super consensus book, so follow it religiously
will sometimes make the opposite effect on the clearness of code. 

Considering the fact that we have quite a clear code watched by hundreds
of eyes on the open web , we will omit this PR and stay with the current 
style. Thanks again for your effort and time.

PR #394, opened 2016-04-27
Make sure TransactionInfo objects can survive the RLP roundtrip

ewiner:
We've been hitting a bunch of `ArrayIndexOutOfBoundsException` and `RuntimeException("wrong decode attempt")` errors when trying to call `Blockchain.getTransactionInfo()`.  After investigating, it looks like `TransactionInfo.index` doesn't serialize and deserialize properly to RLP encoding.

The call to `RLP.decodeInt()` below was double-decoding the number, since `RLP.decode2()` already recursively traversed the whole data structure and decoded everything.  Since this is a fix on the decoding side, existing transaction stores should be fine to use.


Nashatyrev:
Thanks! Nice catch!

PR #395, opened 2016-04-27
Fix eth_estimateGas() for simple txs

ewiner:
I noticed that `eth_estimateGas()` was returning 0 for everything if you weren't calling a contract; I believe it should return 21000 instead, the standard tx cost.

In at attempt to fix it, I ended up removing some of the code that tries to short-circuit execution for simple value transfers - if I understand this correct, some of that code (e.g. making sure gas usage stays under the gas limit for the transaction) should apply even if you're not actually calling a contract.

I'm not sure if this is the right way to solve the problem, please just let me know if I should try another way.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/jsonrpc/JsonRpc.java
+++ b/ethereumj-core/src/main/java/org/ethereum/jsonrpc/JsonRpc.java
@@ -36,6 +36,7 @@ public interface JsonRpc {
         public String gasPrice;
         public String value;
         public String data; // compiledCode
+        public String nonce;
 
         @Override
         public String toString() {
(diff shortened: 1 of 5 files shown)


ethereumj-core/src/main/java/org/ethereum/jsonrpc/JsonRpc.java
```diff
@@ -36,6 +36,7 @@ public String toString() {
         public String gasPrice;
         public String value;
         public String data; // compiledCode
+        public String nonce;
```
Nashatyrev: What is the meaning of `nonce` here? I couldn't find it in the JSON-RPC docs: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_call

Nashatyrev:
Fixed the problem in a slightly different way: https://github.com/ethereum/ethereumj/commit/5dcfd1b743f9eec563d2a64dae03abf7b2e73e68
Also added the nonce to CallArguments: https://github.com/ethereum/ethereumj/commit/898a2867671c718064969f659f89478d87f1f44f
Thanks for the PR!

ewiner:
Cool, yours seems like a cleaner and less risky implementation.

PR #400, opened 2016-05-19
fix spelling of getSnapshotTo function

jlopp:
Simple spelling fix to bring it in line with other snapshot function names.


Nashatyrev:
Thanks!

Issue #399, opened 2016-05-12
NullPointerException in 1.2.5


jlopp:
When I upgraded EthereumJ to 1.2.5 I'm seeing a lot of these NPEs, though they don't appear to be fatal:

`05-12 09:44:49.756 [nioEventLoopGroup-128-1] ERROR net - P2p handling failed
io.netty.handler.codec.EncoderException: java.lang.NullPointerException
	at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:107)
	at io.netty.handler.codec.MessageToMessageCodec.write(MessageToMessageCodec.java:116)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:633)
	at io.netty.channel.AbstractChannelHandlerContext.access$1900(AbstractChannelHandlerContext.java:32)
	at io.netty.channel.AbstractChannelHandlerContext$AbstractWriteTask.write(AbstractChannelHandlerContext.java:908)
	at io.netty.channel.AbstractChannelHandlerContext$WriteAndFlushTask.write(AbstractChannelHandlerContext.java:960)
	at io.netty.channel.AbstractChannelHandlerContext$AbstractWriteTask.run(AbstractChannelHandlerContext.java:893)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:356)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:357)
	at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:110)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:137)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException: null
	at org.ethereum.net.rlpx.MessageCodec.getCode(MessageCodec.java:206)
	at org.ethereum.net.rlpx.MessageCodec.splitMessageToFrames(MessageCodec.java:168)
	at org.ethereum.net.rlpx.MessageCodec.encode(MessageCodec.java:160)
	at org.ethereum.net.rlpx.MessageCodec.encode(MessageCodec.java:37)
	at io.netty.handler.codec.MessageToMessageCodec$1.encode(MessageToMessageCodec.java:67)
[...]

PR #401, opened 2016-05-25
[npe-fix-attempt] Issue #399, NPE on MessageCodec.getCode()

gubatron:
It seems that due to an unsupported protocol the MessageCodec's messageCodesResolver object
is not initialized. Then on that "hack" method getCodec() it blows up.

Sorry if this not the right fix, seems like a quick workaround though, perhaps the right fix has its roots somewhere at `P2pHandler::setHandshake`, but at least this could silence the NPE's reported on Issue #399


Nashatyrev:
Since this exception is not critical I wouldn't like just to hide it for now. Hopefully we will find time soon to manage this issue.

gubatron:
👍

PR #402, opened 2016-05-26
Develop

mhoudg:
Fix an exception when offset larger than 32


mhoudg:
This may cause an exception
```
java.lang.IllegalArgumentException: 128 > 32
    at java.util.Arrays.copyOfRange(Arrays.java:3519) ~[na:1.8.0_31]
    at org.ethereum.core.CallTransaction$Bytes32Type.decode(CallTransaction.java:320) ~[ethereumj-core-1.2.5-RELEASE.jar:na]
    at org.ethereum.core.CallTransaction$Function.decode(CallTransaction.java:484) ~[ethereumj-core-1.2.5-RELEASE.jar:na]
```

Nashatyrev:
Thanks!

PR #412, opened 2016-06-07
Bytes32Type.decode bug

thedoctor:
Arrays.copyOfRange takes `(array, from, to)`, not `(array, from, length)`


Nashatyrev:
Thanks!

PR #415, opened 2016-06-10
Allow raw transactions to be submitted to `StandaloneBlockchain`

pelle:
I have an application where I need to be able to test submitting regular transactions to `StandaloneBlockchain` as created from an external app.

Currently it only allows you to either sendEther, submit contracts or call contract functions.

There are no unit tests that I can find for StandaloneBlockchain and I'm not enough of a java programmer to make write new ones, but I have local unit tests in clojure that test it and it seems to work. 

I'm absolutely open to any suggestions on improving this. So please be brutal. 😄


Nashatyrev:
Heh, thanks for PR! Just recently did the same :)

PR #429, opened 2016-06-28
Add TransactionExecutionSummary and miner rewards to the onBlock callback

ewiner:
This adds two pieces of information into the `onBlock` callback:

* The `TransactionExecutionSummary` for each tx in the block.
* A `Map<byte[], BigInteger>` of all mining rewards (including uncles & tx fees) paid out in the block. (requested by @Melvillian).

This includes a breaking change to `EthereumListener` - but `EthereumListenerAdapter` (which I assume most people are using instead of the bare Listener) is still backwards compatible.


Nashatyrev:
Cool, thanks for the PR!

PR #432, opened 2016-06-30
Fix the -connect command line arg

ewiner:
I was using the `org.ethereum.Start` main method and found that `SystemProperties` was throwing an error if I used the `-connect` parameter.  This fixes that problem, and adds a `-connectOnly` parameter to set the peer and turn off peer discovery.


Nashatyrev:
Thanks! 
But also keep in mind that other peers can actively connect to you if your peers was running with discovery before and was broadcasting its enode address

PR #433, opened 2016-07-05
Any 0 in MULMOD results directly in 0

nicksavers:
Avoids ```ByteUtil.copyToArray()``` for zero-values. 
Noticed a 10x performance improvement in a basic test-case where ```word1``` was ZERO. 

This could also be done for ```DataWord.mul(DataWord word)```


romanman:
Thanks @nicksavers looks good

PR #435, opened 2016-07-11
Don't force a GC when the Blockchain is flushed to disk

ewiner:
We're managing GC timing pretty tightly in our app and would rather not have this `System.gc()` call every 1000 blocks.  Is it alright to remove?  Or could we make it configurable?


Nashatyrev:
Yep, this `gc()` is also bothering me. But we have a strategy to flush when we are near out of heap memory (config `cache.flush.memory > 0`) and in this case we need to do full gc. I know there should be better solution, but have no resources for it now.
I did the following fix https://github.com/ethereum/ethereumj/commit/dd6371de6f32ea8a00881f683119f4e88281fc2d which I believe would be suitable for you

ewiner:
Yep, https://github.com/ethereum/ethereumj/commit/dd6371de6f32ea8a00881f683119f4e88281fc2d is a better fix than mine.

PR #436, opened 2016-07-11
Indicate in the onBlock callback whether this is the new best block

ewiner:
This adds an `isBestBlock` boolean to the `onBlock()` callback.  This would help us with some logic we'd like to run in our callbacks only when there's a new tip of the blockchain, instead of when a fork is in progress.


Nashatyrev:
@ewiner Can you please describe your use case? 
I just think the `isBest` flag is not too informative and I'm trying to think of more generic indicator 
If e.g. 2 concurrent chain branches are receiving, i.e. blocks #10, #10', #11, #11', #12, #12', ..., all of them could be imported as best. What benefit would you get from this information? 
M.b. `rebranch` indicator would be more feasible?

ewiner:
Sorry for the long delay here.  I have one listener that "reaches back" from the blockchain tip to look for transactions matching some criteria with N confirmations.  If it were to act on a non-best block, those transactions might not actually have at least that level of confirmations.

I could just track total difficulty in my listener though, so this isn't strictly necessary.  Happy to close if you prefer.

Melvillian:
@ewiner. In the latest version of `master` BlockSummary has 'totalDifficulty' set on it (see line 572 of `BlockchainImpl.java`), so you can get it in the `onBlock()` callback and use it from there.

Though sometimes @jlopp and I need to get the total difficulty for a block in places in parts of the code unrelated to `onBlock()`, so for that we patched ethereumj to expose the `BlockStore.getTotalDifficultyForHash()` method, which reaches into LevelDB and gets the total difficulty at an arbitrary block height. So you could try that as well. I'll submit a pull request by mid next week if it will help.

Nashatyrev:
Since the flags you requested are easily calculated from the existing info I think there is no need to add them ti API: 
```
isBest = totalDiff > previousBestTotalDiff
isRebranched = isBest && block.parentHash != previousBest.hash
```
Thanks for suggestion!

PR #439, opened 2016-07-11
Support for custom classloaders

ewiner:
`Class.forName()` can be dangerous inside a library, because various containers, like J2EE, OSGI, Apache Spark, and the Play Framework (which I'm using) use different classloaders for app code and for library code.  So when I tried to use a custom `BlockchainNetConfig` and a custom genesis JSON file, EthereumJ wasn't able to load them.  This change allows the user to pass in a custom classloader, and defaults to the one used to load the `SystemProperties` class.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/crypto/jce/SpongyCastleProvider.java
+++ b/ethereumj-core/src/main/java/org/ethereum/crypto/jce/SpongyCastleProvider.java
@@ -7,10 +7,10 @@ import java.security.Security;
 
 public final class SpongyCastleProvider {
 
-  public static final String PROVIDER_NAME = BouncyCastleProvider.PROVIDER_NAME;
+  private static final String PROVIDER_NAME = BouncyCastleProvider.PROVIDER_NAME;
 
   private static class Holder {
-    private static final BouncyCastleProvider INSTANCE;
+    private static final Provider INSTANCE;
 
     static {
       Provider provider = Security.getProvider(PROVIDER_NAME);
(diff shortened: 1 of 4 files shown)


ethereumj-core/src/main/java/org/ethereum/crypto/jce/SpongyCastleProvider.java
```diff
@@ -3,26 +3,11 @@
...
-
-      if (provider == null) {
-        INSTANCE = new BouncyCastleProvider();
-        Security.addProvider(INSTANCE);
-      } else {
-        assert provider instanceof BouncyCastleProvider;
-        INSTANCE = (BouncyCastleProvider) provider;
-      }
-    }
+    private static final Provider INSTANCE = new BouncyCastleProvider();
```
Nashatyrev: Hi, what was wrong with that class regarding classloader?

dwhjames: Hey @Nashatyrev, this was originally my addition.

We realized later that this pattern of dynamically appending a Java Security Provider causes issues in a JVM environment with hot reloading (dev environments for web development, for example).

`Security.addProvider` adds the object instance to some global lookup table. This results in a space / classloader leak when doing class reloading as a reference to the original classloader is kept, and ClassCastExceptions occur in ECKey code because instances from the SpongyCastle library are crossing classloader instances. This is because `Security.getProvider` will return the instance that came from the original classloader.

I removed the code that interacts with the security provider lookup table, and there is no detriment to EthereumJ. There is no code path where a specific security provider is looked up by string name, rather than explicitly passing an instance of a Provider.

Nashatyrev:
Sounds reasonable. Thanks for explanation!

PR #443, opened 2016-07-14
Upgrade to SpongyCastle 1.53.0.0

nicksavers:
- Upgrade to Spongy Castle dependency version 1.53.0.0
- This results in having to use the KeccakDigest in the RLPx code
- Only use the 4.5x faster Saphir2 implementation of Keccak-256 and Keccak-512 in HashUtil
- Updated remaining classes to only use ```HashUtil.sha3(...)```
- Remove obsolete Sha3Helper class
- Use the latest https://github.com/ethereum/tests commit-id

Tested successfully on the entire chain from main net.

** Notes
Upgrading to SC 1.54.0.0 results in errors in RlpxConnectionTest


ethereumj-core/src/main/java/org/ethereum/net/rlpx/EncryptionHandshake.java
```diff
@@ -360,11 +360,11 @@ public ECPoint getRemotePublicKey() {
...
-        public SHA3Digest getIngressMac() {
+        public KeccakDigest getIngressMac() {
             return ingressMac;
         }
 
-        public SHA3Digest getEgressMac() {
+        public KeccakDigest getEgressMac() {
             return egressMac;
         }
     }
```
Nashatyrev: Hi @nicksavers, thanks for this PR!
Did I get it right, that the latest SC version adjusted SHA-3 to the NIST standard and it is no more Keccak-256 (used in Ethereum) compatible? 
(https://medium.com/@ConsenSys/are-you-really-using-sha-3-or-old-code-c5df31ad2b0#.mnzw52i6b)

nicksavers: That's correct. Even the yellowpaper was updated to reflect this.
It's confusing now that not everybody has renamed it yet, I admit. 

Eventually all references to SHA3 should disappear... even the operation name in the VM, Solidity and Web3.js, ... but I don't think everybody is ready for that yet.

I purposely didn't change the method name `sha3(..)` to `kec(..)` and `kec512(...)` to keep the change less controversial, but I'd be happy to see that change.

nicksavers:
@Nashatyrev note that I added an extra file (`Keccak512.java`) with copyright from project RNRT SAPHIR along the 4 that were already there and a small piece of code to abstract some functions with copyright from somebody else. These aren't the same MIT license as ethereumj. I did not check if these are fully compatible, so this might need to be resolved.

Nashatyrev:
@nicksavers Got you! Thanks again!

PR #444, opened 2016-07-18
Specify ping timer scheduling parameter as long type

jlopp:
I was getting a build error in Eclipse because it was assuming that config.getProperty() was returning an int in this case since it considered the second parameter (value 5) as an int rather than as a long.


Nashatyrev:
Thanks!

PR #447, opened 2016-07-20
upgrade gradle shadow plugin to 1.2.3

splix:
It seems that version 1.2.0 is obsolete, I was getting:

````
* What went wrong:
An exception occurred applying plugin request [id: 'com.github.johnrengelman.shadow', version: '1.2.0']
> Failed to apply plugin [class 'com.github.jengelman.gradle.plugins.shadow.ShadowJavaPlugin']
   > Could not create task of type 'ShadowJar'.
````

### CODE REVIEW

**Comment by Nashatyrev** [comment]
Hi @splix, did you try `shadowJar` (with lowercase `s`) ? This just worked fine for me

**Comment by splix** [comment]
@Nashatyrev it was failing just on grails startup, I mean for every gradle command. And Intellij Idea was unable to import this file. I guess it was some incompatibility with Gradle 2.14.

File: ethereumj-core/build.gradle [hunk 1]
```diff
@@ -4,17 +4,19 @@ buildscript {
...
         classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:3.0.1'
+        classpath "com.github.jengelman.gradle.plugins:shadow:1.2.3"
     }
 }
 
 plugins {
     id 'application'
     id 'jacoco'
-    id 'com.github.johnrengelman.shadow' version '1.2.0'
+    id 'com.github.johnrengelman.shadow' version '1.2.3'
```
Nashatyrev: Hey, @splix!
Doesn't applying only this line change works for you? For me it works ok: the plugin was successfully downloaded from Bintray

splix: no, didn't work too. what version of gradle are you using? my is 2.14

splix: Just checked again on a fresh clone, applying changes to this line only does work for me. Version change is enough

Nashatyrev: @splix Cool! Would you change the PR? I could commit the fix directly in other case

splix: ok, sure

PR #479, opened 2016-08-30
jsontestsuite moved to tests

zilm13:
jsontestsuite and tck moved to test
Running tck from gradle tested by
```
 ./gradlew tckRun  -PfilePath=tests/StateTests/stBlockHashTest.json
```
 from https://github.com/ethereum/tests/
tckRun supports [only state tests](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/tck/RunTck.java#L45) so state test was used.


Nashatyrev:
@zilm13, cool! Thanks for this PR!
Just wondering regarding coverage decrease (-0.8%) after this commit. I thought it shouldn't change. Do you have any ideas?

Nashatyrev:
@zilm13, cool! Thanks for this PR!
Just wondering regarding coverage decrease (-0.8%) after this commit. I thought it shouldn't change. Do you have any ideas?

zilm13:
@Nashatyrev Added test to only added util function, it haven't changed it.
It shows that coverage of 1 untouched class dropped by 1 line, but it shouldn't change whole coverage by almost 1 percent. Really surprised too.

PR #481, opened 2016-09-01
Manually start syncing with peers

Melvillian:
Adds a method to the Ethereum interface for manually kickoff the SyncManager connecting with other peers. This is designed to be used in a scenario where you start your program and perform some logic that makes use of the Ethereum object without it automatically loading blocks from peers, and once that logic is complete, start the sync process.


Nashatyrev:
@Melvillian , do you initially set `sync.enabled = false`, then call 
```
SystemProperties.getDefault().setSyncEnabled(true);
ethereum.initSyncing()
```
?

Melvillian:
@Nashatyrev Yes, that's exactly what we're using.

Nashatyrev:
Merged here: https://github.com/ethereum/ethereumj/commit/9e72f60655d262a53ec3d2f18383288153227dce
Thanks for PR!

PR #491, opened 2016-09-07
Fixing development restart behavior

zilm13:
Let me explain the idea:
Main entry is in EthereumImpl, it catches Spring Context shutdown signal and processes all closes.
The problem is that classes not managed by Spring stays alive and continue their common flow. So we need to integrate their lifecycle into Spring's, so we could manage them on shutdown. And these ThreadGroups should be somehow reinitialised after restart.
Do we really need "prototype" scope for PeerClient?


Nashatyrev:
Hm, nice catch! We don't really need it to be `prototype`!
Can you please then also change `ctx.getBean(PeerClient.class)` to autowired instances. That would look cleaner imho. 
Thanks!

zilm13:
Done.
EthereumImpl.isConnected now looks strange but I've not changed its behaviour, as I understand.

Nashatyrev:
Yep, and it was looking strange before your fix :)

PR #509, opened 2016-09-12
Added config path for Solidity compiler

zilm13:
Haven't find a better way to test static things :(


ethereumj-core/src/main/java/org/ethereum/solidity/compiler/Solc.java
```diff
@@ -25,6 +27,22 @@ private Solc() {
     }
 
     private void init() throws IOException {
+        SystemProperties defaultProps = SystemProperties.getDefault();
+        if (defaultProps != null && defaultProps.customSolcPath() != null) {
+            solc = new File(defaultProps.customSolcPath());
+            solc.setExecutable(true);
```
Nashatyrev: Do you think this call is necessary. Shouldn't it generate exception normally? (I mean solc normally should have root modification privilege)

zilm13:
Sorry, fixed

ethereumj-core/src/main/java/org/ethereum/solidity/compiler/Solc.java
```diff
@@ -25,6 +27,21 @@ private Solc() {
     }
 
     private void init() throws IOException {
+        SystemProperties defaultProps = SystemProperties.getDefault();
```
Nashatyrev: Can you please also make `SolidityCompiler` a Spring @Component to pick up the `SystemProperties` from the AppContext? We eliminated all the static references to it recently.

zilm13:
I've started to do it, and found that it will change whole SolidityCompiler usage, which is used by users and is not an internal-only component. So we will need some facade to provide user SolidityCompiler instance. I think that this should be postponed until we provide new API to provide user a good alternative and not just broke old code.

Nashatyrev:
I did the following pattern: 
```
SolidityCompiler {
  @Autowired 
  SystemProperties config = SystemProperties.getDefault(); 

  public static SolidityCompiler getInstance() {
    ....
  }
}
```
Would it be fine to keep compatibility?

Nashatyrev:
In internal components (like `JsonRpcImpl`) you may autowire `SolidityCompiler`

zilm13:
@Nashatyrev you will have two instances of both Solc and SolidityCompiler in that case, 1) with defaultConfig and 2) with wired one. When used as static you will get (1), when used as Autowired Bean, you'll get (2)
If it's ok, I could refactor it.

Nashatyrev:
@zilm13 I think that's fine for now. Just make sure you initialize static instance lazily

PR #512, opened 2016-09-13
Add ECDSASignature toHex() method

csdodd:
Converts the signature to a hex representation. There is some discussion on how the v value should be treated.
v can currently be a value of 0, 1, 27, or 28 (00, 01, 1b, 1c, in hex respectively). This method normalises v to be either 00, or 01.
The reasoning for this is that geth seems to also output with values in this range.


csdodd:
You can verify this method using the following page: https://etherchain.org/verify/signature

Generate a hex address as follows:
```
    public String sign(final ECKey ecKey, final String hexString) {
        try {
            final byte[] msgHash = sha3(TypeConverter.StringHexToByteArray(hexString));
            final ECKey.ECDSASignature signature = ecKey.sign(msgHash);
            return signature.toHex();
        } catch (final Exception e) {
            e.printStackTrace();
        }
        return null;
    }
```

local.properties
```diff
@@ -0,0 +1,11 @@
...
+# Do not modify this file -- YOUR CHANGES WILL BE ERASED!
+#
+# This file must *NOT* be checked into Version Control Systems,
+# as it contains information specific to your local configuration.
+#
+# Location of the SDK. This is only used by Gradle.
+# For customization when using a Version Control System, please read the
+# header note.
+#Tue Sep 13 09:24:40 CEST 2016
+sdk.dir=/Users/colin/Library/Android/sdk
```
Nashatyrev: @csdodd Many thanks for this PR! 
Just one thing: can you please remove this file?

PR #515, opened 2016-09-13
Unify genesis loading with other ethereum clients

zilm13:
I don't like length < 2, but Geth is just comparing it this way too, not considering that there could be not 0.


Nashatyrev:
`data.length() < 2`: looks strange since `0xa` -> `0a` but `a` -> `00`
M.b. geth assumes the string should start with `0x`? Then this check makes sense

zilm13:
May be jusr change it to "0" check? So we will be compatible in most way with Geth and it will not look so strange. But still there will be duch strange thing:
0 -> []
0x0 -> [0]

zilm13:
Here's  Geth helper:
```Go

func FromHex(s string) []byte {
	if len(s) > 1 {
		if s[0:2] == "0x" {
			s = s[2:]
		}
		if len(s)%2 == 1 {
			s = "0" + s
		}
		return Hex2Bytes(s)
	}
	return nil
}
```

Nashatyrev:
Yep, Geth looks odd. 
I believe both `0x0` and `0` should be `[0]`, but `0x` should be `[]`

ethereumj-core/src/main/java/org/ethereum/util/ByteUtil.java
```diff
@@ -603,12 +603,16 @@ public static short bigEndianToShort(byte[] bs, int off) {
...
      * Converts string hex representation to data bytes
+     * Accepts following hex:
+     *  - with or without 0x prefix
+     *  - with no leading 0, like 0xabc -> 0x0abc
      * @param data  String like '0xa5e..' or just 'a5e..'
      * @return  decoded bytes array
      */
     public static byte[] hexStringToBytes(String data) {
-        if (data == null) return EMPTY_BYTE_ARRAY;
+        if (data == null || data.length() < 2) return EMPTY_BYTE_ARRAY;
```
Nashatyrev: I would just omit this change

ethereumj-core/src/main/resources/genesis/olympic-messy.json
```diff
@@ -0,0 +1,64 @@
+{
```
Nashatyrev: Could you move this file to `test` folder?

zilm13:
I agree with you
0 and 0x0 should be handled the same way

ethereumj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -21,6 +21,9 @@
  */
 public class BlockHeader {
 
+    public static int NONCE_LENGTH = 8;
+    public static int HASH_LENGTH = 32;
+    public static int ADDRESS_LENGTH = 20;
```
Nashatyrev: Please make fields `final`

ethereumj-core/src/main/java/org/ethereum/core/genesis/GenesisLoader.java
```diff
@@ -66,9 +66,9 @@ public static Genesis loadGenesis(SystemProperties config, InputStream genesisJs
     }
 
 
-    private static Genesis createBlockForJson(GenesisJson genesisJson){
+    private static Genesis createBlockForJson(GenesisJson genesisJson) throws Exception {
```
Nashatyrev: Do we need this checked exception here?

Nashatyrev:
Cool! Thanks!

PR #522, opened 2016-09-16
Fixed ArrayIndexOutOfBounds error when performing parseEvent

martintreurnicht:
Currently if you parse an event log that contains indexed parameters, it fails with an ArrayIndexOutOfBounds error. This is because indexed parameters are included as topic and not into the event log data field

As per https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI#events, i've implemented a fix for this. I've tested it and it seems to work without any issues


Nashatyrev:
Thanks for PR!

PR #523, opened 2016-09-17
Fixed wrong receipt hash for some blocks (due to duplicated LogInfo)

zilm13:
Tested on all 3 networks, passed full chains correctly.


ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -455,7 +455,6 @@ public void createContract(DataWord value, DataWord memStart, DataWord memSize)
 
         track.commit();
         getResult().addDeleteAccounts(result.getDeleteAccounts());
-        getResult().addLogInfos(result.getLogInfoList());
```
zilm13: There was merge one screen above.
Actually merge also adds deleted accounts but looks like `addDeleteAccounts` line should stay there. There's still some magic for me there.

PR #525, opened 2016-09-19
Add way to load genesis from other locations

### LINKED ISSUE #514: Add ability to specify arbitrary file for genesis in config
Currently you can set the resource file only

cupuyc:
Resolve #514

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
@@ -751,11 +751,7 @@ public class SystemProperties {
 
     @ValidateMe
     public String genesisInfo() {
-
-        if (genesisInfo == null)
-            return config.getString("genesis");
-        else
-            return genesisInfo;
+        return genesisInfo == null ? config.getString("genesis") : genesisInfo;
     }
 
     @ValidateMe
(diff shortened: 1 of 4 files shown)


Nashatyrev:
Hi @cupuyc, honestly I don't like such 'duality' i.e. if you have 'frontier.json' in the current directory it would have 
1. backward incompatible behavior 
2. no way to specify the usual logic (i.e. load from resources) 

Let's think on other options... M.b. add `genesisFile` config param which just have higher priority over `genesis` parameter if exist?

cupuyc:
@Nashatyrev My points are:
 - keep config file simpler/shorter;
 - I think we don't need backward compatibility. Because loading config from resource directory is not something, that out users will use. Configs are usually placed into separate directory;
 - for tests we continue to use loading from resource directory or can use trick like: 
```
        URL url = GenesisLoadTest.class.getClassLoader().getResource("genesis/frontier-test.json");
        // full path
        loadGenesis(url.getPath());
```
Then I think, that users should NOT know about loading genesis from resources directory.

Please make final decision including my thoughts.

Nashatyrev:
Hm, but the aim of `genesis` param was primarily to load one of predefined genesis files in the `/genesis` resource directory, so users normally specify it this way (i.e. `frontier.json` or `morden.json`). 
By backward compatibility I mean that if some user has a file named `frontier.json` in his working dir this file will be loaded after your fix instead of the resource file it should be loaded from (when running with default config file)

cupuyc:
@Nashatyrev As I told, my vision is to not expose loading genesis from resources for generic users. So that only core developers will meet with inconvenience if any.

Please make decision on how it should be implemented and write final word.

PR #526, opened 2016-09-19
Feature: broadcast received pending txs to other peers

zilm13:
I'm not sure if we should broadcast only NEW_PENDING txs, have not found any info about this.


Nashatyrev:
Looks good! Did you tested this in some way?

zilm13:
Duplicate #551

PR #527, opened 2016-09-19
Feature: distribute new blocks received from the peers

zilm13:
Maybe also exclude peers that are too far behind us both for new blocks and new transactions? Say, 1000 or 192 blocks behind. There is a small chance anyone is keeping queue longer than 1000 blocks.


ethereumj-core/src/main/java/org/ethereum/sync/SyncManager.java
```diff
@@ -267,6 +267,12 @@ private void produceQueue() {
 
                 if (syncDone && (importResult == IMPORTED_BEST || importResult == IMPORTED_NOT_BEST)) {
                     if (logger.isDebugEnabled()) logger.debug("Block dump: " + Hex.toHexString(wrapper.getBlock().getEncoded()));
+
+                    // Propagate block to the net after successful import
+                    if (wrapper.isNewBlock()) {
+                        Channel receivedFrom = channelManager.getActivePeer(wrapper.getNodeId());
+                        channelManager.sendNewBlock(wrapper.getBlock(), receivedFrom);
+                    }
```
Nashatyrev: Can you please submit this logic to a separate async executor? 
This code doesn't seem to do any blocking stuff in the current implementation, but I'd prefer to invoke any network related stuff asynchronously from this method just for the case.

zilm13:
Duplicate #551

PR #530, opened 2016-09-22
fix encodeInt for values > 2^255

thedoctor:
`encodeInt` causes an array out of bounds exception for any value > 2^255 since `toByteArray` returns the 2's complement representation (which requires 257 bits – 33 bytes for that case). To avoid this, we input the value whose 2's complement is the desired unsigned value when bigInt > 2^255.


Nashatyrev:
Oh, we have the code duplication here with `CallTransaction` inner classes :/

thedoctor:
Sweet 👍

PR #533, opened 2016-09-27
Feature/457 leveldb on mac and repair

### LINKED ISSUE #457: Autorepair corrupted leveldb
After some experiments I got my LevelDB database corrupted, probably because by stopping the process during flush. When first time it happened I just resynced db from the beginning. 

Then it happened second time, and I decided to a fix in the code that tries to repair database when it got exception from leveldb driver. That worked fine for me and I didn't notice any problem after that. 

I'd like to add it as a PR, but there're few moments to discuss:

1. This probably requires a config option, and it's not clear for me should it be enabled by default or not
2. I don't think there it a way to make any real world tests for this scenario, just because "db corruption" can be anything. So what about tests?
3. Repaired after corrupted db may contain invalid data. That probably means that it requires full db revalidation, from block 0 (or tx 0, depends actually on which db was corrupted). 
4. later can be done in background or foreground, or maybe it should require a special config or command line option. I have no idea. Maybe it isn't required at all

PS That was a quick fix for my situation so obviously I don't have any configs or tests at this moment

cupuyc:
Resolve #457 with:
 - auto repair database once if any error occurs during database initialization;
 - switch to leveldb-1.18 on Mac;
 - adjust `BuildInfo.java` to not throw error when file missing. It is annoying to do `gradle build` after branch switch.

And yes, DB reset must be required for Mac. For me once it worked fine, but with second test DB detected corruption, repaired and then met "No parent" error.

### DIFF
--- a/ethereumj-core/build.gradle
+++ b/ethereumj-core/build.gradle
@@ -11,6 +11,7 @@ buildscript {
     }
 }
 
+
 plugins {
     id 'application'
     id 'jacoco'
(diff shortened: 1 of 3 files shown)


Nashatyrev:
Cool, thanks!

PR #536, opened 2016-09-27
Switch to logback instead of log4j. Async file appender is used to im…

### LINKED ISSUE #451: Move core default logger from log4j to logback
Logback appears to be more flexible with more appender implementations.
The target is to put by default rolling archiving file appender so the logs size don't exceed some reasonable threshold

cupuyc:
Resolve #451 with:
 - replaced log4j with **logback 1.1.7**;
 - updated `BasicSample` with logback logic;
 - updated `AdvancedDeviceUtils` with logback logic;
 - used **async appender** to improve performance. Added shutdownHook option to ensure logs are flushed before JVM shutdown;
 - configured **zip** encoding for rolled files via logback tools.


ethereumj-core/src/main/java/org/ethereum/samples/BasicSample.java
```diff
@@ -85,18 +91,27 @@ public BasicSample(String loggerName) {
...
+        consoleAppender.setContext(lc);
+        consoleAppender.start();
+
+        ch.qos.logback.classic.Logger logger = (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(loggerName);
+        logger.addAppender(consoleAppender);
+        logger.setLevel(Level.DEBUG);
+        logger.setAdditive(false); /* set to true if root should log too */
+
+        return logger;
     }
```
Nashatyrev: The idea behind tweaking logger settings was to start the logger as usual then switch console appender to output only sample logger messages and set ERROR level for others
When running `BasicSample` I still see all other loggers INFO messages

PR #537, opened 2016-09-27
Fixed skip behavior of Eth62 getHeaders

zilm13:
Actually Geth tries to return something for hash not from the main chain but:
- From [Wire Protocol descirption](https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol):
`beginning at block block (denoted by either number or hash) in the canonical chain`
- Every header from not main chain will violate protocol description
- Moving towards leafs could be a random challenge
- Currently we throw out hash and return part of canonical chain, so first block of response and request hash doesn't match
- It's allowed to return blocks less than requested

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/BlockHeader.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/BlockHeader.java
@@ -342,6 +342,7 @@ public class BlockHeader {
 
     private String toStringWithSuffix(final String suffix) {
         StringBuilder toStringBuff = new StringBuilder();
+        toStringBuff.append("  hash=").append(toHexString(getHash())).append(suffix);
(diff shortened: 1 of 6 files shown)


ethereumj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -342,6 +342,7 @@ public String toString() {
 
     private String toStringWithSuffix(final String suffix) {
         StringBuilder toStringBuff = new StringBuilder();
+        toStringBuff.append("  hash=").append(toHexString(getHash())).append(suffix);
```
zilm13: Please, I wanted it for several times.

Nashatyrev:
@zilm13 Did you tested changes on a real network?

zilm13:
both on real network and with isolated geth node
but not full run, just 20k blocks
I could run full test tonight

PR #548, opened 2016-09-29
add totalDifficulty property to Block

Melvillian:
`totalDifficulty` exists on the BlockSummary which is passed to the `onBlock()` callback, but not on the BlockSummary's Block object. In many cases it is easier to work with and pass via function calls the Block object rather than the more hefty BlockSummary, but Block doesn't have a totalDifficulty associated with it. This is useful for when you're trying to compare two blocks to see which one is the best block (i.e. which one has greater total difficulty).

This PR adds `totalDifficulty` as a gettable and settable property on the Block so totalDifficulty can easily be passed through various functions in apps that use the ethereumJ library.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/Block.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Block.java
@@ -46,6 +46,8 @@ public class Block {
 
     private Trie txsState;
 
+    private BigInteger totalDifficulty; // summed difficulty of this block and all its ancestors
+
 
     /* Constructors */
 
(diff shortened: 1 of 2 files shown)


Nashatyrev:
Hi @Melvillian, 
The Block class canonically reflects the Ethereum spec, so I would prefer not adding any other stuff here. 
If you check here https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java#L355-355 you would see that using a wrapper internally to store info directly not related to Block. 
I'd suggest you to use a wrapper or a Map to store totDiff and any other Block related stuff you need

PR #551, opened 2016-09-30
Feature/broadcast new blocks 507

zilm13:
Broadcast new txs
Broadcast new blocks/hashes
Resend txs on disconnect
Test


zilm13:
Removed #526 #527 as duplicates

ethereumj-core/src/main/java/org/ethereum/sync/SyncManager.java
```diff
@@ -204,7 +235,7 @@ private void blockRetrieveLoop() {
                         // to get more chances to receive block body promptly
                         for (BlockHeaderWrapper blockHeaderWrapper : bReq.getBlockHeaders()) {
                             Channel channel = pool.getByNodeId(blockHeaderWrapper.getNodeId());
-                            if (channel != null) {
+                            if (channel != null && channel.isIdle()) {
```
zilm13: Sometimes cause peer drop with reputation reduce.
Actually we have no guarantee in multi-thread env on the next few lines but had not ever run it in.

ethereumj-core/src/main/java/org/ethereum/sync/SyncManager.java
```diff
@@ -269,10 +316,61 @@ private void produceQueue() {
...
+                    logger.error("Error processing unknown block", e);
+                }
+            }
+        }
+    }
+
+    /**
+     * Processing new blocks from queue
+     */
+    private void newBlocksDistributeLoop() {
```
Nashatyrev: M.b. let's move blocks and tx broadcasting logic from `SyncManager` to let's say `ChannelManager`. 
Those functions has actually few in common with syncing and more with networking

ethereumj-core/src/main/java/org/ethereum/sync/SyncManager.java
```diff
@@ -91,9 +104,8 @@ public void accept(BlockWrapper blockWrapper) {
     private CountDownLatch receivedHeadersLatch = new CountDownLatch(0);
     private CountDownLatch receivedBlocksLatch = new CountDownLatch(0);
 
-    private Thread syncQueueThread;
-    private Thread getHeadersThread;
-    private Thread getBodiesThread;
+    // All workers-like thread loops to simplify shutdown
+    private List<Thread> threadLoops = new ArrayList<>();
```
Nashatyrev: I would also return back these fields to simplify further class reading. (each of these threads has its own essential destiny :) )

zilm13:
@Nashatyrev please, take a look!

Nashatyrev:
Cool, thanks!

Issue #406, opened 2016-06-01
Process inbound Eth requests without locking BlockchainImpl


Nashatyrev:
Requests like `GetBlockHashaes`, 'GetBlockBodies` are now executed under `BlockchainImpl` lock so many threads are frequently blocked by such calls.

zilm13:
I've made such test and don't see that it's blocking:
```java
/**
 * Testing whether Eth handler is blocking BlockchainImpl
 */
public class EthBlockingTest {

    @Configuration
    @NoAutoscan
    public static class SysPropConfig1 {
        static Eth62 testHandler = null;

        @Bean
        @Scope("prototype")
        public Eth62 eth62() {
            return testHandler;
        }

        static SystemProperties props = new SystemProperties();

        @Bean
        public SystemProperties systemProperties() {
            return props;
        }
    }

    @Test
    public void testTest() throws FileNotFoundException, InterruptedException {

        SysPropConfig1.props.overrideParams(
                "peer.listen.port", "30334",
                "peer.privateKey", "3ec771c31cac8c0dba77a69e503765701d3c2bb62435888d4ffa38fed60c445c",
                "genesis", "genesis-light.json",
                "database.dir", "testDB-1");

        final Blockchain blockchain = new BlockchainImpl(SysPropConfig1.props) {
            @Override
            public synchronized List<BlockHeader> getListOfHeadersStartFrom(BlockIdentifier identifier, int skip, int limit, boolean reverse) {
                return null;
            }

            @Override
            public synchronized boolean isBlockExist(byte[] hash) {
                return true;
            }
        };
        
        SysPropConfig1.testHandler = new Eth62(SysPropConfig1.props, blockchain, new CompositeEthereumListener()) {
            protected void sendMessage(EthMessage message) {
                try {
                    System.out.println("Start sleeping...");
                    Thread.sleep(10000);
                    System.out.println("Stop sleeping...");
                } catch (InterruptedException e) {
                    e.printStackTrace();
                }
            }
        };

        final AtomicBoolean result = new AtomicBoolean();
        result.set(false);

        synchronized (this) {
            ExecutorService executor1 = Executors.newSingleThreadExecutor();
            executor1.submit(new Runnable() {
                                 @Override
                                 public void run() {
                                     SysPropConfig1.testHandler.processGetBlockHeaders(new GetBlockHeadersMessage(1L, 192));
                                 }
                             }
            );
            this.wait(1000);
[...]

PR #561, opened 2016-10-06
Remove locking possibility of eth handler by blockchainImpl

zilm13:
It's fix for #406 
I had not moved synchronized methods, Eth handler was using to Blockstore, because there are a lot of not "storing" logic in them, I think this approach is better.
Also I'm not sure I did correct with synchronized in BlockStore. Marked all methods that are reading and writing changing objects.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
+++ b/ethereumj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
@@ -51,7 +51,7 @@ public class IndexedBlockStore extends AbstractBlockstore{
         }).withCacheSize(256);
     }
 
-    public Block getBestBlock(){
+    public synchronized Block getBestBlock(){
 
         Long maxLevel = getMaxNumber();
         if (maxLevel < 0) return null;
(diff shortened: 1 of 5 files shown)


ethereumj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -91,7 +91,7 @@ public void flush(){
 
 
     @Override
-    public void saveBlock(Block block, BigInteger cummDifficulty, boolean mainChain){
+    public synchronized void saveBlock(Block block, BigInteger cummDifficulty, boolean mainChain){
```
Nashatyrev: Could you please synchronize all public methods?

Nashatyrev: Especially flush :)

Nashatyrev:
@footloosejava Yep this is basically not the best pattern, but here it is made intentionally for now, since there were issues when accessing internals during block import so users were recommended to do the stuff under `BlockchainImpl` lock. 
This doesn't seem to be a 'vulnerability' as the code is basically open for any kind of manipulations by the client code. 
`init()` method was called only during AppContext initialization, so no issue here

Issue #497, opened 2016-09-08
Stop trying importing when 'No Parent' error 


Nashatyrev:
'No Parent' means that something went really wrong when importing a block from the main chain and this error may be assumed unrecoverable

Melvillian:
We are getting this error all the time, usually after a "DBException - Closed" error from get org.ethereum.datasource.LevelDbDataSource.get, or from a mismatch between the Block receipt, logBloom, or state variables. We have a monit process set up to watch for such log lines and reboot the process if they appear, but ethereumJ is close to unusable when it silently stops working with these "No parent" errors

Nashatyrev:
@Melvillian I hope we will shortly get rid of `DBException - Closed` problems after fixing https://github.com/ethereum/ethereumj/issues/427
We also working now on https://github.com/ethereum/ethereumj/issues/431
Let's get back to your problems after resolving those issues. It's pity to hear you get those errors constantly. On our servers they are not appearing mostly (with a couple of rare exceptions)

cupuyc:
@Nashatyrev For me, `No parent` error happened when I switch networks and use same database directory. Can we protect database in that case?

Nashatyrev:
@cupuyc Good proposal. We may think on this protection later

PR #563, opened 2016-10-06
Checking DB match with genesis on block load

zilm13:
Quit if our db doesn't match genesis
Regarding https://github.com/ethereum/ethereumj/issues/497#issuecomment-246594272


Nashatyrev:
Yeah, that's better than nothing. But this still doesn't prevent from e.g. ETH/ETC db's conflicts

PR #564, opened 2016-10-07
Feature/337 remove fake info in discovery

### LINKED ISSUE #337: With discovery mode enabled a new node id is used for each connection
Having the discovery mode enabled (which is the default setting) causes ethereumj to always use a new nodeid when connecting to a node. This causes problems with ethernodes.org as single nodes are counted multiple times.

I think this line is the root cause of the issue: https://github.com/ethereum/ethereumj/blob/f389f28bb816ed0a74c69c14188aa7b451ac971b/ethereumj-core/src/main/java/org/ethereum/net/server/Channel.java#L100

cupuyc:
To resolve #337 
 - fixed Node rlps encoding + test;
 - using proper remote port for incoming connections;
 - removed fake nodeId and wrong port during discovery;
 - added sample class with 3 peers in same JVM, interacting with each other.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
+++ b/ethereumj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
@@ -49,7 +49,7 @@ public class HandshakeHandler extends ByteToMessageDecoder {
     private static final Logger loggerNet = LoggerFactory.getLogger("net");
 
     private FrameCodec frameCodec;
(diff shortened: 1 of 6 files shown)


cupuyc:
I also noticed warning:
`16:43:26.079 WARN [net]  Can't decrypt AuthInitiateMessage from /127.0.0.1:55398. Most likely the remote peer used wrong public key (NodeID) to encrypt message.`
I ignored it as task goal was achieved.

PR #566, opened 2016-10-10
Added 'maven' plugin to main build

footloosejava:
This will generate pom.xml artifacts for using in maven project
importing.


Nashatyrev:
Thanks!

PR #569, opened 2016-10-13
Handle correctly discovery endpoint with fake nodeId

### LINKED ISSUE #567: PrivateNetworkDiscoverySample discovery issues
When running `PrivateNetworkDiscoverySample` there 2 unresolved discovery issues:
1. 
```
WARN [sampleNode-2]  Discovery keeps silence. Waiting more...
```
The `NodeManager` has all 3 nodes however one of them missed corresponding `onNodeDiscovered` callback
2. 
```
WARN [net]  Can't decrypt AuthInitiateMessage from /127.0.0.1:51674. Most likely the remote peer used wrong public key (NodeID) to encrypt message.
```
That means some node was initially trying to connect another node with wrong `PeerID`. Later all peers appeared connected with each other

cupuyc:
To resolve #567 with changes:
 - we have special handling for discovery endpoints nodes.  They are not stored in all nodes table and are not used for TCP connections;
 - discovery endpoint node could become normal node once we receive it's real nodeId via discovery;
 - home node is always included to NEIGHBORS response in discovery;
 - we don't dispatch home node via `ethereumListener.onNodeDiscovered()`

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
+++ b/ethereumj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
@@ -73,7 +73,7 @@ public class HandshakeHandler extends ByteToMessageDecoder {
     public void channelActive(ChannelHandlerContext ctx) throws Exception {
         channel.setInetSocketAddress((InetSocketAddress) ctx.channel().remoteAddress());
         if (remoteId.length == 64) {
-            channel.setNode(remoteId);
+            channel.initWithNode(remoteId);
             initiate(ctx);
         } else {
             handshake = new EncryptionHandshake();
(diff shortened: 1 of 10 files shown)


ethereumj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
```diff
@@ -252,6 +256,8 @@ private void decodeHandshake(final ChannelHandlerContext ctx, ByteBuf buffer) th
 
                 final HelloMessage inboundHelloMessage = (HelloMessage) message;
                 channel.setNode(remoteId, inboundHelloMessage.getListenPort());
+                // use node into discovery
+                nodeManager.getNodeHandler(channel.getNode());
```
Nashatyrev: I suppose the line above `setNode()` should do the same (i.e. `getNodeHandler` should be invoked)
Can we remove this then?

ethereumj-core/src/main/java/org/ethereum/net/rlpx/Node.java
```diff
@@ -103,6 +105,14 @@ public Node(byte[] rlp) {
         this.id = idB;
     }
 
+    /**
+     * @return true if this node is endpoint for discovery loaded from config
+     */
+    public boolean isDiscoveryNode() {
+        // discovery node has nodeId with doubled size
+        return id.length > 64;
+    }
```
Nashatyrev: Honestly I don't like the idea of making ID 128 bytes length. All code (including user) relies on 64 bytes and can be broken at any unpredictable place. 
Can we identify Node by its host(IP)+port+nodeId (i.e. at least change the `equals` method) and set the fake `0x00 x 64` ID? Or at least set the first 32 bytes to zero for fake ID? I'd prefer the first

Nashatyrev: > explicitly have new field isDiscoveryEndpoint;
Cool, just give it a more general name like `hasNodeId` or `isFakeNodeId`

PR #570, opened 2016-10-13
Not closing peer discovery db if not initialized

andrewla:
Is PERSIST is not configured for peer discovery, then the db is never created, so this close is guaranteed to throw an NPE on shutdown


Nashatyrev:
Thanks!

PR #576, opened 2016-10-17
Eip 150 block target fix

willcrozi:
Take into account the additional 6000 block delay introduced to allow for extra testing before hard-fork.


willcrozi:
Oops didn't see the identical pull request! Closing now.

PR #579, opened 2016-10-20
Improvements to ABI parsing

martintreurnicht:
* Improved fault tolerance when parsing ABI json.
* Added field `payable`
* Added Function Type `fallback`


Nashatyrev:
Thanks for PR!

PR #591, opened 2016-11-09
EIP155, EIP160 implementation

zilm13:
Let's start
I don't like mess with 13/null in Transaction, but I don't want to expose 13 to the users. Maybe use null instead?
Also need more tests, will research how to adopt jsontestsuite transaction tests for it.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/config/blockchain/Eip160HFConfig.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/blockchain/Eip160HFConfig.java
@@ -1,18 +1,45 @@
 package org.ethereum.config.blockchain;
 
 import org.ethereum.config.BlockchainConfig;
+import org.ethereum.core.Transaction;
+import org.ethereum.vm.GasCost;
 
 /**
- * Created by Anton Nashatyrev on 14.10.2016.
+ * Hard fork includes following EIPs:
+ * EIP 155 - Simple replay attack protection
+ * EIP 160 - EXP cost increase
+ * EIP 161 - State trie clearing (invariant-preserving alternative)
  */
 public class Eip160HFConfig extends Eip150HFConfig {
+
+    static class GasCostEip160HF extends GasCostEip150HF {
+        public int getEXP_BYTE_GAS()        {     return 50;     }
+    }
+
+    private static final GasCost NEW_GAS_COST = new GasCostEip160HF();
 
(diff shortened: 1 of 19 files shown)


ethereumj-core/src/test/java/org/ethereum/jsontestsuite/suite/JSONReader.java
```diff
@@ -115,7 +115,7 @@ public static String getTestBlobForTreeSha(String shacommit, String testcase){
 
     public static List<String> getFileNamesForTreeSha(String sha){
 
-        String result = getFromUrl("https://api.github.com/repos/ethereum/tests/git/trees/" + sha);
+        String result = getFromUrl("https://api.github.com/repos/ethereum/tests/git/trees/" + sha + "?recursive=1");
```
Nashatyrev: What was wrong with this?

ethereumj-core/src/main/java/org/ethereum/config/blockchain/Eip160HFConfig.java
```diff
@@ -1,18 +1,45 @@
...
+
+    @Override
+    public Integer getChainId() {
+        return 1;
+    }
+
+    @Override
+    public boolean acceptTransactionSignature(Transaction tx) {
+        if (!super.acceptTransactionSignature(tx)) return false;
+        return getChainId().equals(tx.getChainId());
```
Nashatyrev: `        return tx.getChainId() == null || getChainId().equals(tx.getChainId());`

zilm13: Yeah, this one would be better.

ethereumj-core/src/test/java/org/ethereum/jsontestsuite/GitHubVMTest.java
```diff
@@ -150,4 +160,28 @@ public void testRandomVMGitHub() throws ParseException {
...
+                );
+
+        for (String fileName : fileNames) {
+            if (!includedFiles.contains(fileName)) continue;
+            System.out.println("Running: " + fileName);
+            String json = JSONReader.loadJSONFromCommit(fileName, shacommit);
+            GitHubJSONTestSuite.runGitHubJsonVMTest(json);
+        }
+
+    }
```
Nashatyrev: All those tests are executed by GitHubVMTest. We just need to change commit hash to the right one

PR #599, opened 2016-11-16
Fixed difficulty encoding in Block

zilm13:
According to the Ethereum White Paper, difficulty is scalar value, so this kind of encoding should be used for it. PR includes test which fails for old style.
ps: Not sure that I'm using the right base for it.


Nashatyrev:
Looks good. Can you please rebase this fix for develop? 
Anyways it was not for 1.3.x, sorry for misunderstanding

PR #603, opened 2016-11-21
Improve `eth_call` with support of `pending` state

cupuyc:
Implemented that by copying part of `EthereumImpl` code to JsonRpc. I don't think we need a method in facade at the moment, because I don't see any nice way to implement it and document friendly.


Nashatyrev:
I though the JsonRpc impl was completely moved to Harmony and is ready to be removed from core?

PR #608, opened 2016-11-24
Removed Parity disable for fastsync due to fix in Parity

zilm13:
I see Parity 1.3.13 and 1.4.4 latest releases (they maintain two trees) in active nodes
1.4.4 have fix, 1.3.13 not
1.4.3 have fix too (this was the first release from 1.4 tree with HF implementation)

I've tried to leave only Parity nodes and NODE_DATA's were retrieving with list of peers only of Parity's, so I think this line could be removed.


Nashatyrev:
I saw a lot of ERROR exceptions in the log when running with non-fixed peers. 
Can you please check this with 1.3.13 peers and process them gently (without ERRORs in log) and ban and disconnect such peers?

PR #614, opened 2016-11-29
Feature/fast sync parallel

zilm13:
Need to finish with some issues, but any feedback would be appreciated


zilm13:
Implemented:
- parallel download of node_data and blocks/headers
- correct node data saving to maintain number of node usage
- improved headers download with split
- several other fixes

Please, review

ethereumj-core/src/main/java/org/ethereum/sync/SyncQueueImpl.java
```diff
@@ -244,16 +271,30 @@ private boolean addHeaderPriv(BlockHeaderWrapper header) {
...
+            startNumber = maxNum + 1;
         } else {
             List<HeaderElement> longestChain = getLongestChain();
-            if (rnd.nextBoolean()) {
-                return new HeadersRequestImpl(longestChain.get(longestChain.size() - 1).header.getNumber(), MAX_CHAIN_LEN, false);
-            } else {
-                return new HeadersRequestImpl(longestChain.get(longestChain.size() - 1).header.getNumber(), MAX_CHAIN_LEN, true);
-            }
+            startNumber = longestChain.get(longestChain.size() - 1).header.getNumber();
+            if (!rnd.nextBoolean()) reverse =  true;
```
Nashatyrev: When requesting reverse headers to close a gap no need to request `count` headers it makes sense to request only `MAX_CHAIN_LEN`

ethereumj-core/src/main/java/org/ethereum/sync/SyncQueueImpl.java
```diff
@@ -244,16 +271,30 @@ private boolean addHeaderPriv(BlockHeaderWrapper header) {
...
         }
+
+        int headersCount;
+        if (endBlockNumber != null) headersCount = (int) Math.min(count, endBlockNumber - startNumber + 1);
+        else headersCount = count;
```
Nashatyrev: Can you please adjust formatting either with `{}` or replace by `?:`

ethereumj-core/src/main/java/org/ethereum/sync/SyncQueueImpl.java
```diff
@@ -35,6 +35,20 @@ public HeadersRequestImpl(byte[] hash, int count, boolean reverse) {
         private boolean reverse;
 
         @Override
+        public List<HeadersRequest> split(int maxCount) {
+            if (this.hash != null) return Collections.<HeadersRequest>singletonList(this);
+            List<HeadersRequest> ret = new ArrayList<>();
+            while(count > 0) {
+                int reqSize = Math.min(maxCount, count);
+                ret.add(new HeadersRequestImpl(start, reqSize, reverse));
+                count -= reqSize;
```
Nashatyrev: This class is kinda immutable, can you please use local var for `count`?

ethereumj-core/src/main/java/org/ethereum/sync/SyncQueueImpl.java
```diff
@@ -35,6 +35,20 @@ public HeadersRequestImpl(byte[] hash, int count, boolean reverse) {
...
+                if (reverse) start -= reqSize;
+                else start += reqSize;
```
Nashatyrev: Bad if-else java formatting. Can we use `?:` here?

ethereumj-core/src/main/java/org/ethereum/sync/SyncQueueIfc.java
```diff
@@ -40,6 +42,11 @@
     HeadersRequest requestHeaders();
```
Nashatyrev: I believe we can remove this method then (for interface clarity)

ethereumj-core/src/main/java/org/ethereum/sync/FastSyncManager.java
```diff
@@ -191,27 +184,19 @@ public TrieNodeRequests(TrieNodeRequest request) {
...
+        TrieNodeRequest request = pendingNodes.get(hash);
+        if (request.requestSent.isEmpty()) pendingNodes.remove(hash);
     }
 
     synchronized void processTimeouts() {
         long cur = System.currentTimeMillis();
-        for (TrieNodeRequests nodeRequests : new ArrayList<>(pendingNodes.values())) {
-            for (TrieNodeRequest request: nodeRequests.requests) {
+        for (TrieNodeRequest request : new ArrayList<>(pendingNodes.values())) {
+            synchronized (request.requestSent) {
```
Nashatyrev: That's why `ConcurrentHashMap` made it work (it internally synchronizes on its instance)

ethereumj-core/src/main/java/org/ethereum/sync/FastSyncManager.java
```diff
@@ -154,6 +154,10 @@ public synchronized void reqSent(Long requestId) {
             requestSent.put(requestId, timestamp);
         }
 
+        public synchronized Set<Long> requestIdsSnapshot() {
```
Nashatyrev: To avoid sync problems with `requestSent` and eliminate additional locks you need to synchronize here on a `FastSyncManager` instance:
```
synchronize(FastSyncManager.this) {
}
```

ethereumj-core/src/main/java/org/ethereum/sync/FastSyncManager.java
```diff
@@ -154,6 +154,10 @@ public synchronized void reqSent(Long requestId) {
             requestSent.put(requestId, timestamp);
```
Nashatyrev:
```
synchronize(FastSyncManager.this) {
}
```

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth62.java
```diff
@@ -792,6 +792,13 @@ public synchronized void dropConnection() {
         disconnect(USELESS_PEER);
     }
 
+    @Override
+    public synchronized boolean setStatus(SyncState syncState) {
+        if (!sentHeaders.isEmpty() && !headerRequests.isEmpty()) return false;
```
zilm13: @Nashatyrev I did it that way first time, but we need to return peer to IDLE if we locked it but there is no need to use in the end, so I added such checking instead

Issue #622, opened 2016-12-08
1.4.0 Conflict on block #2687389


Nashatyrev:
```
17:16:08.484 INFO [sync]  Success importing BEST: block.number: 2687388, block.hash: 291801, tx.size: 3, time: 0.013s
17:16:08.491 WARN [blockchain]  Block's given Receipt Hash doesn't match: 9d5c65424a596c204c67733c5c04f1ced49b51769be94c3f451c67f2da80a9a0 != e2ab7099ef9b2e803b728a73c5beae6ac0f9978fbd6f0885a7a6ed7be846a393
17:16:08.491 WARN [blockchain]  Calculated receipts: [TransactionReceipt[
  , postTxState=1592934b4075bf89c788bc2d5586acb7227748586ca7318c284137d6b740d9b8
  , cumulativeGas=044dec
  , gasUsed=044dec
  , error=
  , executionResult=
  , bloom=00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
  , logs=[]], TransactionReceipt[
  , postTxState=d57d159c73ce529a4ec238b72f040c3948b05939769521e65b2a82878237637b
  , cumulativeGas=049ff4
  , gasUsed=5208
  , error=
  , executionResult=
  , bloom=00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
  , logs=[]], TransactionReceipt[
  , postTxState=ffafa5f4c5f2a92224ce701aa6438622b4fbbb3d488aaf0e4e8253e5d420fe80
  , cumulativeGas=04f1fc
  , gasUsed=5208
  , error=
  , executionResult=
[...]

PR #625, opened 2016-12-12
Fixed: some touched accounts were not verified for eip161 deletion

zilm13:
Let's clarify
https://github.com/ethereum/EIPs/issues/161

An account changes state when:
1. it is the target or refund of a SUICIDE operation for zero or more value; 
1. it is the source or destination of a CALL operation or message-call transaction transferring zero or more value; 
1. it is the source or newly-creation of a CREATE operation or contract-creation transaction endowing zero or more value; 
1. as the block author ("miner") it is recipient of block-rewards or transaction-fees of zero or more. 

Looks like we could answer *YES* on all of this, but you'd better check too. So:
- Do we now check and merge results of all cases?
- Should we move touched accounts to ProgramResult?

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
+++ b/ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
@@ -464,8 +464,6 @@ public class Program {
 
         if (!byTestingSuite())
             track.commit();
-        getResult().addDeleteAccounts(result.getDeleteAccounts());
-        getResult().addTouchAccounts(result.getTouchedAccounts());
 
         // IN SUCCESS PUSH THE ADDRESS INTO THE STACK
         stackPush(new DataWord(newAddress));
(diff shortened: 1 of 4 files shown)


zilm13:
@Nashatyrev should fix #622

Nashatyrev:
>    Do we now check and merge results of all cases?
I added merge for regular call but forgot about create. Do you mean any other cases? 

>    Should we move touched accounts to ProgramResult?
Yep probably. Also it is possible that we shouldn't add touched accounts in case of throw. But if this will work on live then it actually doesn't matter much, since all accounts are cleared now and this execution branch will not be executed anymore

zilm13:
> I added merge for regular call but forgot about create. Do you mean any other cases?
Just asking you to check it too according to specs to clarify that we didn't forgot anything else.
> Also it is possible that we shouldn't add touched accounts in case of throw.
Yeah, that's why I've asked about ProgramResult. So I will refactor it a bit now.

zilm13:
Please, check

ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -465,6 +465,7 @@ public void createContract(DataWord value, DataWord memStart, DataWord memSize)
         if (!byTestingSuite())
             track.commit();
         getResult().addDeleteAccounts(result.getDeleteAccounts());
+        getResult().addTouchAccounts(result.getTouchedAccounts());
```
Nashatyrev: M.b. make it a part of `ProgramResult.merge()` ?

zilm13:
@Nashatyrev Updated to use merge

PR #633, opened 2016-12-15
Fix NullpointerException when using zero gas configuration for transactions

martintreurnicht:
When creating private chain where gas is switched off, there is some NullpointerExceptions that pop up all over the place, these changes make sure that those don't occur.

I've also cleaned up some of the access logic for `transaction` properties, because these have complex accessor logic it really should be private and have setters. One `NullpointerException` was caused by gasPrice property being accessed directly rather than using getter


martintreurnicht:
@Nashatyrev The tests that are failing seem to be failing even before my changes came into affect. I'll attempt to fix them but really these should've been fixed by whom ever broke them originally

PR #638, opened 2016-12-19
Switch to solc 0.4.7

cupuyc:
To fix #638 
 - use solc 0.4.7;
 - fix json parsing unknown properties.

I see some related tests are failed before my changes. I hope merging solc update will not break it fuse.
```
ImportLightTest.operateNotExistingContractTest
TransactionTest.multiSuicideTest
LockBlockchainTest.testStatus (looks unrelated to solidity)
```


Nashatyrev:
Thanks!

Issue #631, opened 2016-12-15
Optimization: Transaction.validate() needs to be called on wire inbound ounly


Nashatyrev:
`Transaction.validate()` is pretty heavy operation due to `getSender()` call. 
We need to validate Tx when receiving it from wire only, when reading Tx from DB validation is obsolete

Nashatyrev:
Also we have serious performance regression in `PendingStateImpl`: most inbound txs are filtered by their hash and there is no need to validate them again.

Nashatyrev:
Fixed with PR

PR #639, opened 2016-12-19
Validate transaction only when we are unsure with data safety

zilm13:
@Nashatyrev Regarding #631 
Validating only when we are not sure about data safety + in tests


zilm13:
Please take a look on such implementation
Validating tx's in pendingState only if not duplicated
Plus test added

PR #650, opened 2016-12-21
Fixed blockQueue size grows over limit

### LINKED ISSUE #645: SyncManager.blockQueue grows far above limit
With limit 2000 with regular sync `blockQueue` can grow up to 8000 which causes OOM on `-Xmx1G` on heavy attacker blocks

zilm13:
Should fix #645


zilm13:
@Nashatyrev checked on fastsync too
everything ok

PR #655, opened 2016-12-22
Feature/geneis config for hive

cupuyc:


ethereumj-core/src/main/java/org/ethereum/net/rlpx/Node.java
```diff
@@ -53,10 +53,19 @@ public Node deserialize(DataInput in, int available) throws IOException {
     // discovery endpoint doesn't have real nodeId for example
     private boolean isFakeNodeId = false;
 
-    public static Node createWithoutId(String address) {
-        final ECKey generatedNodeKey = ECKey.fromPrivate(sha3(address.getBytes()));
+    public static Node createWithoutId(String addressOrEnode) {
```
Nashatyrev: Can you please rename that method then (enode:// has ID) and add some docs

PR #657, opened 2016-12-23
Feature/genesis config and hive changes

cupuyc:
Reopened prev PR here, because last branch became dirty.
 - allow to config blockchain via `genesis.json`;
 - adjust genesis to handle values in different formats;
 - allow to set boot node in enode format;
 - allow manual blocks importing like geth does in RLP format.


ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -784,7 +790,7 @@ public boolean minerStart() {
     @ValidateMe
     public byte[] getMinerCoinbase() {
```
cupuyc: Allow coinbase in hex with `0x` or without

ethereumj-core/src/main/java/org/ethereum/manager/BlockLoader.java
```diff
@@ -3,8 +3,8 @@
```
cupuyc: Allow manual import of RLP encoded blocks

ethereumj-core/src/main/java/org/ethereum/net/rlpx/Node.java
```diff
@@ -53,10 +53,24 @@ public Node deserialize(DataInput in, int available) throws IOException {
     // discovery endpoint doesn't have real nodeId for example
```
cupuyc: Allow boot node in enode format

PR #659, opened 2016-12-25
add option to create contracts by compiled json

UrsZeidler:
A simple extension to allow the deployments of compiled contracts.


UrsZeidler:
Ups.

PR #663, opened 2016-12-26
Tracks no successful import for timeout seconds

zilm13:
@Nashatyrev how about this for tracking "No parent"? 10 minutes should be enough even for attacker blocks and test will stop one time at least. But if restart will let sync goes right we will not catch it with such handling


zilm13:
10 minutes looks too much. How long does it take to import hacker/cleaning block?

PR #666, opened 2016-12-26
Fixing #665 fail processing incoming GetNodeData

zilm13:
Maybe move this logic to Blockchain?

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
+++ b/ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
@@ -7,7 +7,7 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.ethereum.config.SystemProperties;
 import org.ethereum.core.*;
 import org.ethereum.db.BlockStore;
-import org.ethereum.db.RepositoryRoot;
+import org.ethereum.db.StateSource;
 import org.ethereum.listener.CompositeEthereumListener;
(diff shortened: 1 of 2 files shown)


ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
```diff
@@ -40,7 +41,7 @@
 
     private static final EthVersion version = V63;
 
-    @Autowired
+    @Autowired @Qualifier("defaultRepository")
```
Nashatyrev: Nope, that wouldn't work. `defaultRepository` is just a bean for initializing `BlockchainImpl` (probably not the best solution). You may do `BlockchainImpl.getRepository()` here

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
```diff
@@ -95,8 +96,9 @@ protected synchronized void processGetNodeData(GetNodeDataMessage msg) {
         );
 
         List<Value> states = new ArrayList<>();
+        RepositoryRoot repositoryRoot = ((RepositoryRoot) repository);
         for (byte[] stateRoot : msg.getStateRoots()) {
```
Nashatyrev: Could you please also rename `stateRoot(s)` to something like `nodeKey` or `nodeHash` to avoid confusion

zilm13:
Please, check

PR #673, opened 2016-12-28
fix parameter switch

UrsZeidler:
I have switch the parameters of the call. A Simple fix.


UrsZeidler:
I introduced this error with my previous merge request. (I think thats why the build was broken, but I could read the output to figure it out.)

PR #674, opened 2016-12-29
Adding support for the payable keyword to the solidity Abi class

thedoctor:
This PR follows the strategy in https://github.com/ethereum/ethereumj/pull/579 for adding the payable keyword to the ABI parsing mechanism (also a currently-unused `fallback` enum type).

Should be pretty straightforward, except maybe one thing: I'm not sure the custom ObjectMapper settings are a good idea – on one hand it may cause the parser to accept invalid ABIs, hiding errors that would otherwise be noticed quickly, but on the other it prevents this class from breaking entirely if new properties are added to Solidity.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/solidity/Abi.java
+++ b/ethereumj-core/src/main/java/org/ethereum/solidity/Abi.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.Predicate;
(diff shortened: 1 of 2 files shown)


Nashatyrev:
Thanks for the PR!
`DEFAULT_MAPPER` looks good to me. 
The only thing I'm concerned about is wether this change is backward compatible, i.e. would it work with older `solc` versions which have no `payable` property? 
Could you please add another testcase with ABI without `payable`?

ethereumj-core/src/main/java/org/ethereum/solidity/Abi.java
```diff
@@ -146,6 +153,7 @@ public String formatSignature() {
         @JsonCreator
         public static Entry create(@JsonProperty("anonymous") boolean anonymous,
                                    @JsonProperty("constant") boolean constant,
+                                   @JsonProperty("payable") boolean payable,
```
Nashatyrev: Not sure this would work with older solc output

ethereumj-core/src/test/java/org/ethereum/solidity/AbiTest.java
```diff
@@ -0,0 +1,38 @@
...
+        assertEquals(abi.size(), 1);
+
+        Entry onlyFunc = abi.get(0);
+        assertEquals(onlyFunc.type, Type.function);
+        assertEquals(onlyFunc.inputs.size(), 1);
+        assertEquals(onlyFunc.outputs.size(), 1);
+        assertTrue(onlyFunc.payable);
+        assertTrue(onlyFunc.constant);
+    }
+
```
Nashatyrev: Please add a testcase without `payable` property

Nashatyrev:
Just checked, seems it works fine without 'payable' property as well

thedoctor:
Sorry I didn't see the req for changes – my notifications are all messed up. Thanks for merging!

PR #679, opened 2017-01-09
Ip in PingMessage and PongMessage encoded according to specs (4 bytes…

zilm13:
Specs: https://github.com/ethereum/devp2p/blob/master/rlpx.md

test7 in RLPXTest.java contains strange data (from node = byte[0])
I'm not sure, it's a correct RLP

This fix should fix Parity interaction in discovery


Nashatyrev:
Cool, nice catch!

PR #682, opened 2017-01-09
Ensure that pivot block available from majority of peers

zilm13:
Hi.
What do you think about enhancing strategy of choosing pivot. I think it could help in fast sync on chains with long forks. Also it adds some verification for more common cases.
Don't like Pair thing in method return but parent method definitely needed to be decomposed.


Nashatyrev:
Looks good, thanks!

PR #684, opened 2017-01-10
Occupy not all peers on headers download in Fast Sync

### LINKED ISSUE #678: Fast sync doesn't follow up chain when loading headers
I noticed several times next behavior:
First Harmony loads state nodes and synced up to last block;
Then Harmony doesn't update best block for few minutes. At this time it loads headers.
Finally it shows correct best blocks.

2ecc2ax4yc

zilm13:
Should fix #678


Nashatyrev:
Did you check if this fix works fine with a single peer?

zilm13:
@Nashatyrev this line handles this:
https://github.com/ethereum/ethereumj/blob/fix/fast-header-sync-678/ethereumj-core/src/main/java/org/ethereum/sync/HeadersDownloader.java#L90
For me it occupied all peers for header download when there were 2 active peers and leaves one as IDLE with, say, 5

Issue #681, opened 2017-01-09
Longer forks on testnet can stuck sync 


Nashatyrev:
We were targeting the max length (192) of a fork chain since it is very unlikely to happen on a real network. While on a public test network (Ropsten e.g.) with quite low difficulty forks can be of any arbitrary length (recently we had 11K blocks fork on Ropsten). 
The max length is managed by the `SyncQueueImpl` which is keeping all the headers in memory so we can't just increase the max length of a possible fork

Nashatyrev:
Longer forks is also constrained with prune depth, i.e. with 192 depth pruning there is no ways to switch to another fork which starts from e.g. `currentBlock - 1000`

PR #699, opened 2017-01-17
Allow 'reset' to a specific block instead of back to genesis

ewiner:
This was a solution I whipped up for #681 and implemented with success on my Ropsten nodes, which were all stuck on the wrong side of a long fork.  It probably only worked because I had database pruning disabled, but I thought I should post it back here in case it's useful to you.  Feel free to close if it's not a good idea.

I wasn't sure how to detect if the old state had been pruned without doing a whole trie dump, so I left it as a `TODO` in the code.


Nashatyrev:
Thanks! That seems to be useful!
I will fix your TODO item (just need to check if the block state root exists in the Trie)

Issue #605, opened 2016-11-23
Sync status is always "syncing" when node is alone or low blocks count


cupuyc:
When testing private networks I found 2 cases when in ethereumJ core `syncManager.isSyncDone()` is always false:

1. When I run single node without peers with no DB

1. When I run node with 1 peer and have synced all 220 blocks

In described cases geth returns `eth.syncing=false` properly. While ethereumJ returns 
```
{
  "startingBlock": "0x0",
  "currentBlock": "0x0",
  "highestBlock": "0x0"
}
```

I used `develop` version. And noticed similar behavior in `master`.

Nashatyrev:
> When I run single node without peers with no DB
Not sure what `syncManager.isSyncDone()` should return in this case... 
Will try to reproduce the second case on the latest `develop`

Nashatyrev:
Couldn't reproduce the second case with `PrivateMinerSample`: 
- Run sample, wait for several blocks to be mined 
- Stop and remove `sampleDB-2` dir (the regular node db)
- Start again, the regular node is synced up and starts issuing txs (triggered on `onSyncDone`)
- Tried with different chain lengths

cupuyc:
I checked with your scenario.

We discussed on meeting drawback if current sync done logic. Sync done event will only happen  when all blocks downloaded and next block received from network. According that spec all works fine.

@Nashatyrev 
1. Both my issues depend on case when there is no active miner present. I propose to take no actions for my issues as this is non standard scenario.

2. I propose not to use `sync.enabled=false` on mining nodes. This might confuse end users, which copy past samples.

If there are few miners - newly added miner should sync current state first and only then start mining.

zilm13:
should be resolved with #1094

PR #712, opened 2017-01-27
Feature/v64 draft

zilm13:
@Nashatyrev 
Not for merge. For review and discussion. DL part.
Here goes my issues with it:
1) parOffset is 33. Why? I don't know and whether it could be changed. ETH ends on 28
2) Often syncDone is not fired in SyncManager for a long time after unsecure sync
3) What validations I could add to this?
4) Is not final, I will run more chain validation tests


Nashatyrev:
> parOffset is 33. Why? I don't know and whether it could be changed. ETH ends on 28

Probably left from the time when Warp was positioned as Eth64
> Often syncDone is not fired in SyncManager for a long time after unsecure sync

How many top blocks imported before `onSyncDone` fired?

Nashatyrev:
> What validations I could add to this?

The best validation is that you finally get the right state root

zilm13:
@Nashatyrev 
1) If so, did I correctly implement it? It could broke things if there will be, say [eth, par, shh]
2) From my observation it was not fired even after 10 minutes after reaching HEAD when there are 3 peers, and was fired when there are a lot of peers.

Nashatyrev:
> If so, did I correctly implement it? 

Hope so :) 

2. We have an issue on this I couldn't reproduce: https://github.com/ethereum/ethereumj/issues/605
Can you please submit logs there when this happens?

PR #720, opened 2017-02-02
Feature/par1

zilm13:
Implementation of Parity warp sync protocol. Not for merge
Docs:
https://github.com/ethcore/parity/wiki/Warp-Sync
https://github.com/ethcore/parity/wiki/Warp-Sync-Snapshot-Format
@Nashatyrev need code-review
Still there are several issues, but everything is in place.


ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -733,7 +733,7 @@ public void setSyncEnabled(Boolean syncEnabled) {
 
     @ValidateMe
     public boolean isFastSyncEnabled() {
-        return this.syncEnabled == null ? config.getBoolean("sync.fast.enabled") : syncEnabled;
+        return isSyncEnabled() && config.getBoolean("sync.fast.enabled");
```
Nashatyrev: Can you please push this fix to develop?

ethereumj-core/src/main/java/org/ethereum/net/rlpx/FrameCodec.java
```diff
@@ -159,7 +159,7 @@ public void writeFrame(Frame frame, OutputStream out) throws IOException {
             updateMac(ingressMac, headBuffer, 0, headBuffer, 16, false);
 
             dec.processBytes(headBuffer, 0, 16, headBuffer, 0);
-            totalBodySize = headBuffer[0];
+            totalBodySize = headBuffer[0] & 0xFF;
```
Nashatyrev: Please push this fix to develop as well

ethereumj-core/src/main/java/org/ethereum/net/rlpx/MessageCodesResolver.java
```diff
@@ -34,12 +36,21 @@ public void init(List<Capability> caps) {
...
+
             if (capability.getName().equals(Capability.ETH)) {
                 setEthOffset(offset);
                 EthVersion v = fromCode(capability.getVersion());
                 offset += EthMessageCodes.values(v).length;
             }
 
+            if (capability.getName().equals(Capability.PAR)) {
+                // TODO: Why???
+                setParOffset(33);
```
Nashatyrev: Hmm, yep, bad hardcode :/ 
But it seems there is nothing we can do here
Parity doesn't support other protocols like SHH and BZZ so it seems they don't care much about this

ethereumj-core/src/main/java/org/ethereum/net/rlpx/MessageCodesResolver.java
```diff
@@ -34,12 +36,21 @@ public void init(List<Capability> caps) {
...
+                ParVersion v = ParVersion.fromCode(capability.getVersion());
+                // TODO: need we?
+                offset += ParMessageCodes.values(v).length;
```
Nashatyrev: I think it doesn't matter here since `PAR` and (`SHH` and/or `BZZ`) are mutually exclusive

zilm13:
Checked:
1) EthereumJ warp syncs from Pairty's
2) EthereumJ 1 creates a snapshot, EthereumJ 2 syncs from EthereumJ 1
To do:
3) Parity syncs from EthereumJ. Will check it tonight.

zilm13:
Successfully synced Parity from EthereumJ.
Finished.

ethereumj-core/src/main/java/org/ethereum/net/rlpx/MessageCodesResolver.java
arkpar: Message IDs are assigned according to the spec at https://github.com/ethereum/wiki/wiki/%C3%90%CE%9EVp2p-Wire-Protocol#payload-contents
33 comes from ETH/63 currently having 17 messages with IDs 0x10 - 0x20 and that `ETH` < `PAR` when compared in alphabetic order. Note that IDs should are assigned at session negotiating time for mutually supported subprotocols only.

zilm13: @arkpar as you are here, at https://github.com/ethcore/parity/wiki/Warp-Sync it's written that "include snapshot_hash and snapshot_number which signify the block hash and number respectively", but hash is hash of the manifest RLP, and you are validating it by comparison with manifest's RLP hash in Parity's code. Am I correct?

arkpar: Yes, I've fixed the documentation, thank you.

mkalinin:
@zilm13 is there any final decision on this feature?

zilm13:
@mkalinin I don't think we should merge it but next protocol improvements may contain some parts of this feature, so I want to keep branch but no merge

mkalinin:
ok, great! So, let's just close pull request and keep the branch. agree?

zilm13:
@mkalinin agree

PR #742, opened 2017-02-23
Feature/hive simulation tests

cupuyc:
There were a lot of changes.
Let's have a call if you have questions.


cupuyc:
This commit has been reverted. 
https://github.com/ethereum/ethereumj/commit/3f01b5e3d454ba66ccfe3281eaf5dfe0357f05d5
Need to re apply.

ethereumj-core/src/main/java/org/ethereum/manager/BlockLoader.java
```diff
@@ -99,19 +99,22 @@ public void accept(Block block) {
...
                 Path path = Paths.get(fileSrc);
                 // NOT OPTIMAL, but fine for tests
                 byte[] data = Files.readAllBytes(path);
                 RLPList list = RLP.decode2(data);
                 for (RLPElement item : list) {
                     Block block = new Block(item.getRLPData());
                     exec1.push(block);
+                    if (block.getNumber() % 10000 == 0) {
+                        dbFlushManager.flushSync();
+                    }
```
Nashatyrev: This flush shouldn't be there. The flush is now performed automatically and asynchronously on write cache fill. Synchronous flush here will slow import process (and is not good for perf measuring) 
Any reasons behind it?

cupuyc: My test failed until I put this.

Nashatyrev: Didn't the final `flushSync()` do the trick?

cupuyc: Must work. Will remove this intermediate one

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth62.java
```diff
@@ -517,58 +520,52 @@ protected synchronized void sendNextHeaderRequest() {
...
-            for (Pair<Long, byte[]> constraint : constraints) {
-                if (constraint.getLeft() <= getBestKnownBlock().getNumber()) {
-                    blockHashCheck.put(constraint.getLeft(), constraint.getRight());
+            validatorMap = Collections.synchronizedMap(new HashedMap());
+            List<Pair<Long, BlockHeaderValidator>> validators = config.getBlockchainConfig().
+                    getConfigForBlock(blockNumber).headerValidators();
+            for (Pair<Long, BlockHeaderValidator> validator : validators) {
+                if (validator.getLeft() <= getBestKnownBlock().getNumber()) {
+                    validatorMap.put(validator.getLeft(), validator.getRight());
+                    sendGetBlockHeaders(validator.getLeft(), 1, false);
```
Nashatyrev: Didn't you check by the way if this works with more than 1 constraint?

cupuyc: No, separate issue must be created for this as I referenced in description (reverted commit).

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
```diff
@@ -51,6 +51,7 @@
 
     private long connectedTime = System.currentTimeMillis();
     private long processingTime = 0;
+    private long lastReqSentTime;
```
Nashatyrev: What does this for?

ethereumj-core/src/main/java/org/ethereum/config/blockchain/AbstractDaoConfig.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Miners should include marker for initial 10 blocks. Either "dao-hard-fork" or ""
+     */
+    @Override
+    public byte[] getExtraData(byte[] minerExtraData, long blockNumber) {
+        if (blockNumber >= forkBlockNumber && blockNumber < forkBlockNumber + EXTRA_DATA_AFFECTS_BLOCKS_NUMBER ) {
+            if (supportFork) {
+                return DAO_EXTRA_DATA;
+            } else {
+                return new byte[0];
```
Nashatyrev: M.b. `extradata` should be returned here?

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
cupuyc: This is from reverted commit (about more than one hash checks).
I left it as next task will be to bring this fix back.

ethereumj-core/src/main/java/org/ethereum/config/blockchain/AbstractDaoConfig.java
cupuyc: `DAO_EXTRA_DATA` or nothing for the first 10 blocks according to spec.

Nashatyrev: As far as I understand those who were not supporting DAO fork could just not upgrade their clients and as result those blocks have any data but empty: http://gastracker.io/block/1920000
With this check ETC would not work

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
Nashatyrev: I mean this field is declared in superclass. This field is obsolete

ethereumj-core/src/main/java/org/ethereum/config/blockchain/AbstractDaoConfig.java
cupuyc: All fine with ETC
Checked with Harmony `./gradlew runClassic -PuseMavenLocal -Dsync.fast.enabled=th=42f99236037586df16c5b3d1fa66d90b6c85efcaf8edf09bd4cc8d15221f3eef -Ddatabase.dir=database-fast1 -Dlogs.keepStdOut=true`

Our validator doesn't require empty extradata. See: https://github.com/ethereum/ethereumj/pull/742/files#diff-cd718fcb89cfba5123b3c55015aabe44

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth63.java
cupuyc: Got you now. I've mis seen class name. Removed.

Issue #382, opened 2016-04-19
StandaloneBlockchain: hardcode the difficulty to ONE


adridadou:
This is my test code:

```
String contract =
                "contract myContract {" +
                        "  int i1;" +
                        "  function myMethod() returns (int) {" +
                        "    return 23;" +
                        "  }" +
                        "}";

        EthAddress address = ethereumProvider.publishContract(contract);
```

And the publish function does:

```
SolidityContract result = blockchain.submitNewContract(code);
        blockchain.createBlock();
        return EthAddress.of(result.getAddress());
```

and it hangs on "blockchain.createBlock()"
It seems the exact line where it hangs is 
Ethash.getForBlock(b.getNumber()).mineLight(b).get();

Thank you !

Nashatyrev:
@adridadou Please check you have this config modification code:
https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/samples/StandaloneBlockchainSample.java#L42
It might be just a high difficulty calculated for the mined block and it mines much longer

adridadou:
It seems that way. It works much better now. Would it be possible to hardcode the difficulty to ONE for the StandaloneBlockchain ?

Nashatyrev:
Sounds reasonable, but I didn't do it since the setting is changed globally and might affect other instances. But I think I can handle this in some way. 
Thanks for reporting

adridadou:
I see what you mean. Maybe it would be a good idea to hardcode the config but also to have this Standalone class only on the class jar so that people won't use it in their productive code.

Nashatyrev:
Resolved with PR #751

PR #751, opened 2017-03-03
Move net config with minimal difficulty to StandaloneBlockchain

cupuyc:
For #382


adridadou:
so cool! thank you guys

PR #760, opened 2017-03-13
Peer persistence modified to use LevelDB

### LINKED ISSUE #758: Eliminate MapDB usage
Need to refactor discovery for using LevelDB source. MapDB causes a number of failures periodically

zilm13:
Should resolve #758


ethereumj-core/src/main/resources/version.properties
```diff
@@ -1,2 +1,2 @@
 versionNumber='1.5.0'
-databaseVersion=5
\ No newline at end of file
+databaseVersion=6
```
Nashatyrev: Let's leave the DB version as is. The new db is compatible with the previous

PR #762, opened 2017-03-14
Simple approach to fix long insert in quontient filter

zilm13:


zilm13:
I could put MAX_DUPLiCATES to constructor

PR #870, opened 2017-04-19
fixTestInitialInsertTest

arsalankhalid:
Fixed test for initial insert in ProgramMemoryTest
There was a senseless if statement which returned 0 if size <= 0
causing the test to break. 
Unless that if is meant to be there, please advise.


arsalankhalid:
Seems to be a NullPtrException thrown in TransactionStore at TransactionInfo, investigating why this is happening. 

The 'if' statement I removed, wouldn't explain why there's a null pointer being thrown here.

mkalinin:
@arsalankhalid Looks like these changes break `ProgramMemoryTest` execution

Issue #867, opened 2017-04-10
upgrade google guava


adridadou:
Currently, ethj is using guava 16.0.1
The issue is that ethj is using some functions (something about threads) that don't exist anymore in guava 21

As a result, if your project imports ethj + guava 21, then mining silently fails. If you debug, you realize that you get an exception "MethodNotFound" while starting light mining but the application hangs because it happens in a separate thread

jnorthrup:
i would be happy to drive this into an overall push via #878 and see where it leads.  right now i have general sweeping dependency upgrades not complaining relying on the maven versions-plugin to pick the latest and greatest libs and fixing breakage incrementally.

adridadou:
I've created a PR for that

jnorthrup:
which?

PR #871, opened 2017-04-19
NullPointerException when the version is not set

adridadou:
In SystemProperties, there is a default value for the version if it is not set but the order is wrong and we get a NPE if the version is null

---
This change is [Reviewable](https://reviewable.io/reviews/ethereum/ethereumj/871)


adridadou:
in addition to that, I've updated Guava, resolving #867

PR #873, opened 2017-04-21
SolidityContractExtendedFunctionality

arsalankhalid:
Refactored and improved StandaloneBlockchainSample. At the moment the class is very difficult to understand, contains very few comments, and many internal classes. I have tried my best to abstract some of the internals as well as add comments to the flow. This class serves as a great example for beginners to Ethereum and the client, thus it should be well orgaized and very easy to understand.


arsalankhalid:
Anyone have any thoughts on this, some guidance would help.

PR #874, opened 2017-04-21
springConfigUpgrade

arsalankhalid:
Update to Spring libraries, as I'm upgrading the ethereum.properties.
Thought it would be useful to have a separate PR dedicated to fixing Spring version, which are
relevant to adding new db dir in SystemProperties.java while utilizing annotation capabilities.


arsalankhalid:
Ah shoot - I think I know what's causing this, will give it a look.

mkalinin:
@arsalankhalid what is the benefit of this Spring upgrade?

PR #875, opened 2017-04-21
CustomizableDbDirSpringConfig

arsalankhalid:
Creating the functionality to offer the user the ability to set their own database dir from a config file. If no path is set, user home directory will be used.


arsalankhalid:
I know what's causing this, fixing now

zilm13:
There is already configuration for this in etheumj.conf or any other conf you use:

> database {
>     # place to save physical storage files
>     # can be either absolute or relative path
>     dir = database

And you could override it with -Ddatabase.dir="some path" from command line if you want.
What's the purpose of this change?

Sorry for late answer, there were no lib support for some time

PR #878, opened 2017-04-29
towards a gradle distribution with a maven-built library, and some gratuitous upgrades along the way.  

jnorthrup:
there are two gradle builds, the top level, and the ethj codebase below that.

the codebase builds fine with maven though there are some (12) unit tests that do not quite yet run cleanly in intellij ( a goal, a convenience, and something I'm willing to spend effort toward cleaning up towards)  in my fork. 

the upgrades seen fit by automated means in maven were netty, spring, leveldb, spongycastle, and anything with a newer release.  no versions needed to be backed out to infer what the changes were and make those on the spot.

the gradle is tied to the repo and that repo does not talk to all the newer versions in the wild.  having maven build the core and occasionally having a human run the versions upgrade can put less pressure on maintaining said repo.

this work would facilitate a siginficantly more confident set of off-heap refactorings toward the vaunted disruptor pattern execution while potentially removing internally maintained marshalling of json and protocols that can be brought in as apache camel assemblies and maintained more modularly if not completely supplanted.  camel wires up netty in a painless way for instance, so apart from micro optimizations and things that can be done in JH, things like netty can be offloaded to message based codecs and the technical debt can be lowered in the boilerpart areas.  

thanks.

thanks.


jnorthrup:
too noisy.  continuing to refactor.

PR #879, opened 2017-05-02
#706 Make DataWord immutable

vetler:
This makes instances of DataWord immutable - calling public methods to change objects
of this type will no longer change the internal data array, but instead
return new instances of DataWord.

First go at contributing to ethereumj, so feedback with brutal honesty is appreciated!


vetler:
Taking a look at test failures.

vetler:
https://travis-ci.org/ethereum/ethereumj/builds/228120473?utm_source=github_status&utm_medium=notification

PR #885, opened 2017-05-14
Bumps packaged version of solcJ.

MicahZoltu:
`0.4.10` is the latest release and includes `require` and `assert` built-in functions.


mkalinin:
Already stick to [0.4.19](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/build.gradle#L124)

PR #890, opened 2017-05-26
Fix for gitter badge

sandakersmann:
This fixes the bug which causes the image to not be displayed.


nicksavers:
Duplicate of #882

PR #897, opened 2017-06-07
Hotfix/replay block state

mkalinin:


ethereumj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
```diff
@@ -228,35 +228,45 @@ public TransactionReceipt callConstant(Transaction tx, Block block) {
...
                 TransactionReceipt receipt = executor.getReceipt();
-                // TODO: change to repository.getRoot() after RepositoryTrack implementation
-                receipt.setPostTxState(ArrayUtils.EMPTY_BYTE_ARRAY);
+                receipt.setPostTxState(track.getRoot());
                 receipts.add(receipt);
                 summaries.add(summary);
             }
         } finally {
-            repository.rollback();
+            track.rollback();
```
Nashatyrev: Does nothing here (Just FYI, no need to change)

PR #920, opened 2017-08-18
EIP-198 implement a precompile for modular exponentiation

mkalinin:
Implementation of EIP 198: https://github.com/ethereum/EIPs/pull/198

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
+++ b/ethereumj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
@@ -17,11 +17,20 @@
  */
 package org.ethereum.vm;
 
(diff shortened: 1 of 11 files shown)


ethereumj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
```diff
@@ -171,5 +185,116 @@ private static boolean validateV(byte[] v) {
...
+                return adjRes;
+
+            } else {
+                return res;
+            }
+        }
+
+        private long getMultComplexity(long x) {
+
+            long x2 = (long) Math.pow(x, 2);
```
mkalinin: yep, agree on that

mkalinin:

PR #922, opened 2017-08-23
Eip 214

Nashatyrev:
Implementation of https://github.com/ethereum/EIPs/pull/214

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/vm/VM.java
+++ b/ethereumj-core/src/main/java/org/ethereum/vm/VM.java
@@ -20,7 +20,6 @@ package org.ethereum.vm;
 import org.ethereum.config.BlockchainConfig;
 import org.ethereum.config.SystemProperties;
 import org.ethereum.db.ContractDetails;
-import org.ethereum.vm.MessageCall.MsgType;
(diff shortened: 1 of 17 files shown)


ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -974,6 +980,7 @@ else if (oldValue != null && newValue.isZero()) {
                 case LOG3:
                 case LOG4: {
 
+                    if (program.isStaticCall()) throw new Program.StaticCallModificationException();
```
Nashatyrev: Already affected: https://github.com/ethereum/EIPs/pull/214/files

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1288,6 +1302,10 @@ else if (oldValue != null && newValue.isZero()) {
         }
     }
 
+    private void checkStatic(Program program) {
```
mkalinin: Missed that one. This method is not used. do we need to keep it?

mkalinin:

Issue #923, opened 2017-08-25
Byzantium release


mkalinin:
- [x] [EIP-100 Change difficulty adjustment to target mean block time including uncles](https://github.com/ethereum/EIPs/issues/100) | https://github.com/ethereum/ethereumj/pull/916
- [x] [EIP-206 REVERT instruction in the Ethereum Virtual Machine](https://github.com/ethereum/EIPs/pull/206) | https://github.com/ethereum/ethereumj/pull/918
- [X] [EIP-213 Precompiled contracts for addition and scalar multiplication on the elliptic curve alt_bn128](https://github.com/ethereum/EIPs/pull/213) | https://github.com/ethereum/ethereumj/pull/924
- [X] [EIP-212 Precompiled contract for pairing check](https://github.com/ethereum/EIPs/pull/212) | https://github.com/ethereum/ethereumj/pull/933
- [x] [EIP-198 Precompiled contract for bigint modular exponentiation](https://github.com/ethereum/EIPs/pull/198) | https://github.com/ethereum/ethereumj/pull/920
- [x] [EIP-211 New opcodes: RETURNDATASIZE and RETURNDATACOPY](https://github.com/ethereum/EIPs/pull/211) | https://github.com/ethereum/ethereumj/pull/919
- [x] [EIP-214 New opcode STATICCALL](https://github.com/ethereum/EIPs/pull/214) | https://github.com/ethereum/ethereumj/pull/922
- [x] [EIP-649 Metropolis Difficulty Bomb Delay and Issuance Reduction](https://github.com/ethereum/EIPs/pull/669) | https://github.com/ethereum/ethereumj/pull/927
- [x] [EIP-658 Embedding transaction return data in receipts](https://github.com/ethereum/EIPs/pull/658) | https://github.com/ethereum/ethereumj/pull/928
- [X] [EIP-684 Prevent overwriting contracts](https://github.com/ethereum/EIPs/issues/684) | https://github.com/ethereum/ethereumj/pull/932
- [x] Fix `BlockchainTests/TransitionTests` execution
- [x] Add `BlockchainTests/GeneralStateTests` execution
- [x] 100% TCK tests passing

lyless:
[...]

PR #924, opened 2017-08-28
EIP-213 precompiles for BN128 addition and scalar multiplication

mkalinin:


mkalinin:
Implementation of https://github.com/ethereum/EIPs/pull/213

ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -1170,12 +1171,20 @@ public void callToPrecompiledAddress(MessageCall msg, PrecompiledContract contra
...
-            track.commit();
+            if (out.getLeft()) { // success
+                this.refundGas(msg.getGas().longValue() - requiredGas, "call pre-compiled");
+                this.stackPushOne();
+                track.commit();
+            } else {
+                // spend all gas on failure, push zero and revert state changes
+                this.refundGas(0, "call pre-compiled");
+                this.stackPushZero();
+                track.rollback();
```
Nashatyrev: M.b. refactor `PrecompiledContract` a bit and make it similar to regular call, I.e. leave only `execute()` there and throw OOGException if not enough gas for supplied data, and other custom exceptions if invalid data supplied (there you can supply a detailed description on why the call failed) ?

mkalinin: Good idea, but let's make refactoring out of this PR. I spawned related issue https://github.com/ethereum/ethereumj/issues/926 for that

Issue #929, opened 2017-09-01
Message Decoder Conflict


martintreurnicht:
I'm playing around with whisper at the moment and i noticed it gets a `NullPointerException` as soon as whisper is activated, it seems that the `MessageCodec` is where the problem is, it seems that the `frameType`, which is `28` in this case, is both valid for `eth` and `shh`, when resolving the message code from Whisper, it returns `0` which is indicates that it's a `STATUS` message, but it also returns in range for `eth` as `12` (`0x0c`), which indicates as in range, but i can't find a code for it

For now i'm guessing the easiest fix would be to move the whisper decoder above the eth decoder, but i'm not sure if this would cause other errors.

```bash
java.lang.NullPointerException: null
        at org.ethereum.net.eth.handler.Eth63.processNodeData(Eth63.java:193)
        at org.ethereum.net.eth.handler.Eth63.channelRead0(Eth63.java:90)
        at org.ethereum.net.eth.handler.Eth63.channelRead0(Eth63.java:55)
        at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
        at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:108)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
        at io.netty.handler.codec.MessageToMessageCodec.channelRead(MessageToMessageCodec.java:111)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
[...]

PR #931, opened 2017-09-05
Fix Whisper messages being picked up by Eth handler

martintreurnicht:
Simple fix that corrects a error in how offsets are calculated. Changed offsets to apply maximum value + 1 instead of length.

This is needed to resolve https://github.com/ethereum/ethereumj/issues/929

Currently `shh` messages are being picked up by the `eth` message handler. The issue was caused by offsets not calculated correctly, causing whisper messages to be picked up by eth63 handler.

This also seems to resolve the offset hack created for swarm.

Follow up fixes:

* Resolved a concurrency issue where whisper messages were corrupting message data when trying to encode content for sending to multiple peers at the same time.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/net/eth/message/EthMessageCodes.java
+++ b/ethereumj-core/src/main/java/org/ethereum/net/eth/message/EthMessageCodes.java
@@ -17,12 +17,13 @@
  */
 package org.ethereum.net.eth.message;
 
-import org.ethereum.net.eth.EthVersion;
+import static org.ethereum.net.eth.EthVersion.V62;
+import static org.ethereum.net.eth.EthVersion.V63;
 
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


martintreurnicht:
@Nashatyrev If you run my updated `AdaptiveMessageIdsTest` against the current `develop` branch you'll see that `test6` and `test7` will fail with regards to checking the upper boundary for `eth`. This is because of how the offsets are calculated incorrectly for `shh` using the length of the `eth` enum (`0xC`) rather than the max value of the `eth` code (in the case of ETH63 `0x10 + 1`).

This would've worked prior to ETH63 because the enum ordinal and byte value use to be inline, but with ETH63 there is 4 bytes difference between the length and the max value, which is also what was causing the FIXME for calculating the offset for `bzz` in `MessageCodesResolver.init()`

PR #932, opened 2017-09-06
Eip 684: Prevent overwriting contracts

Nashatyrev:
Implementation of https://github.com/ethereum/EIPs/issues/684

Relates #923

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/AccountState.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/AccountState.java
@@ -17,6 +17,7 @@
  */
 package org.ethereum.core;
 
+import org.ethereum.config.BlockchainConfig;
 import org.ethereum.config.SystemProperties;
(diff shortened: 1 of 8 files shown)


ethereumj-core/src/main/java/org/ethereum/core/AccountState.java
```diff
@@ -136,6 +136,10 @@ public AccountState withBalanceIncrement(BigInteger value) {
         return rlpEncoded;
     }
 
+    public boolean isContractExist() {
+        return !FastByteComparisons.equal(codeHash, EMPTY_DATA_HASH) || !BigInteger.ZERO.equals(nonce);
```
mkalinin: Will second condition work as expected if initial nonce > 0?

Nashatyrev: Thanks! Nice catch!

Issue #917, opened 2017-08-17
decodeBigInteger doesn't work properly


tinchou:
Here's a test case:

```
    @Test
    public void encodeDecodeBigInteger() {

        BigInteger expected = new BigInteger("9650128800487972697726795438087510101805200020100629942070155319087371611597658887860952245483247188023303607186148645071838189546969115967896446355306572");
        byte[] encoded = encodeBigInteger(expected);
        BigInteger decoded = decodeBigInteger(encoded, 0);
        assertNotNull(decoded);
        assertEquals(expected, decoded);
    }
```

bendyarm:
I extended my pull request #936  to fix about 29 bugs in RLP including incorporating mlahia's fix.

tinchou:
Thanks for taking a look at this!

I read you fix some 29 bugs, but aren't you adding any tests? It's great to have a correct implementation, but without them you never know when you'll break things again.

bendyarm:
Good point.  I will add tests for some edge cases later next week.

zilm13:
Should be resolved with #975

PR #936, opened 2017-09-22
Issue907: fix encode method to handle Node created from parsedRlp

bendyarm:
Allow encode method to handle case of nondirty Node instantiated by a parsedRlp.

Create a new method makeRlp() that reconstructs and returns an rlp from a
parsedRlp LList object.

Add (parsedRlp != null) case to the if (!dirty) section of the encode method.
Note that the newly created rlp length may be >= 32, so it may need to be saved
and returned as a hash.  The last section of the else clause handles saving a
new rlp.  Move this section of code out of the else clause so it can handle a
new rlp from either the if or else clause.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/trie/TrieImpl.java
+++ b/ethereumj-core/src/main/java/org/ethereum/trie/TrieImpl.java
@@ -115,11 +115,24 @@ public class TrieImpl implements Trie<byte[]> {
             return encode(1, true);
         }
         private byte[] encode(final int depth, boolean forceHash) {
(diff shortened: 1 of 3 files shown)


bendyarm:
Also fix lots of bugs in RLP, including the one from mlahia:
fix #917 : fix bug in decodeBigInteger & refactor encodeElement #940

zilm13:
Should be resolved with #975

zilm13:
@bendyarm I've used part of your code to resolve this and another issues and cannot close PR now. Are you able to do it?
Thank you.

PR #938, opened 2017-09-25
RLP checks

mkalinin:


ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
```diff
@@ -253,8 +253,14 @@ public BytesType() {
 
         @Override
         public byte[] encode(Object value) {
-            if (!(value instanceof byte[])) throw new RuntimeException("byte[] value expected for type 'bytes'");
-            byte[] bb = (byte[]) value;
+            byte[] bb;
+            if (value instanceof byte[]) {
+                bb = (byte[]) value;
+            } else if (value instanceof String) {
+                bb = ((String) value).getBytes();
```
Nashatyrev: Are you sure you want `String.getBytes()` ? M.b. something like hex edcode ?

mkalinin: This extension was inspired by [/ABITests/basic_abi_tests.json](/ethereum/tests/blob/develop/ABITests/basic_abi_tests.json#L10) where `string` literal is converted to `bytes` exactly that way.

Excerpts from Solidity Docs: 
> `string` is equal to `bytes` but does not allow length or index access (for now)
If you want to access the byte-representation of a string `s`, use `bytes(s).length` / `bytes(s)[7] = 'x';`. Keep in mind that you are accessing the low-level bytes of the UTF-8 representation, and not the individual characters!

Therefore, this is a proper conversion in terms of Solidity.

PR #947, opened 2017-11-06
EIP-706 Snappy compression

mkalinin:
Implementation of https://github.com/ethereum/EIPs/pull/706


weisheme:
I took the liberty to take the Python / Go snappy samples from EIP-706 and incorporate into our Java test suite. This way we're certain we can decode Python / Go snappy files.

https://github.com/ethereum/ethereumj/pull/949

PR #949, opened 2017-11-06
Verify that Py/Go Snappy encoded files can be decoded

weisheme:
- EIP-706 provides snappy samples. FrameCodec should be able to decode these samples.


Nashatyrev:
Cool! Thanks a lot!

PR #951, opened 2017-11-08
Remove redundant call to rlpParse()

blmalone:
A proposed change to avoid redundant method invokations. 
Removed redundant call to `rlpParse();` method. It gets called again in `getEncodedRaw()`.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
@@ -257,7 +257,6 @@ public class Transaction {
     }
 
     public byte[] getRawHash() {
-        rlpParse();
         byte[] plainMsg = this.getEncodedRaw();
         return HashUtil.sha3(plainMsg);
     }


ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
```diff
@@ -257,7 +257,6 @@ public boolean isParsed() {
     }
 
     public byte[] getRawHash() {
-        rlpParse();
```
blmalone: this method is always called in the next line at `this.getEncodedRaw();`

blmalone:
@Nashatyrev

mkalinin:
I would not lay on any logic that is invoked in the other method (thus not 100% be completed). All secondary `rlpParse()` invokes are pretty cheap cause they finish [immediately](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java#L191).

blmalone:
@mkalinin cool, thanks.

PR #955, opened 2017-11-15
HashUtil - Reducing cpu and memory

danieleguiducci:
HashUtil: A message digest can be reused to reduce the memory and cpu usage. A MessageDigest is not thread-safe but it can be stored in a ThreadLocal


Nashatyrev:
@danieleguiducci Can you provide any benchmark results? 
I know nothing about `MessageDigest` instance creation footprint

zilm13:
@danieleguiducci  Bad way to go with no performance benefits according to Stackoverflow:
https://stackoverflow.com/a/17555580/8972090
Also you've changed the filename to some strange name with email.
Rejecting.

PR #958, opened 2017-11-24
Upgradejava8

flashboss:
hi. I have an upgrade to java 8 and a refactoring of the unused packages. Can be useful?


zilm13:
Hello, we couldn't merge your PR because of 2 things:
1) Import reorganisation looks not correct and it doesn't follow any good code style guide (for example, [twitter](https://github.com/twitter/commons/blob/master/src/java/com/twitter/common/styleguide.md)) - you've moved static imports to the top, imports from org.ethereum and other packages are messed by order etc. And also there is no sense in import reorganisation until its principles are locked in some kind of checkstyle document which could be used for auto-checking, otherwise it will be broken with further commits and PRs.
2) your PR doesn't pass the tests

zilm13:
Also, I'm going to update EthereumJ to Java8 very soon, so no need to fix this. But if you want, you could setup code styling check for imports and reorganise imports in a good way.

Nashatyrev:
Just upgraded to Java8 in develop
Indeed import reorgs are not subject to be PR'ed, such things heavily mess git history

PR #962, opened 2017-11-30
Upgraded to Java 8

zilm13:
Upgraded to Java 8 because Java 7 will be at the end of the support soon:
- Java 8 Gradle target
- refactored to lambda/method reference where it’s possible
- refactored to comparator usage where it’s possible
- removed org.ethereum.util.Functional and modified all usages of functional classes to use Java 8 native classes instead

Please, take a look @mkalinin , may be I forgot or missed something.
Thank you.


ethereumj-core/src/main/java/org/ethereum/sync/SyncManager.java
```diff
@@ -63,20 +63,14 @@
...
-                        tx.getSender();
-                    }
-                    return blockWrapper;
-                }
-            }, new Functional.Consumer<Throwable>() {
-                public void accept(Throwable throwable) {
-                    logger.error("Unexpected exception: ", throwable);
+            (4, 1000, true, blockWrapper -> {
+                for (Transaction tx : blockWrapper.getBlock().getTransactionsList()) {
+                    tx.getSender();
```
Nashatyrev: Let's party hard :)
```java
(4, 1000, true, blockWrapper -> blockWrapper.getBlock().getTransactionsList().forEach(Transaction::getSender));
```

Nashatyrev:
Java8 rocks!

ethereumj-core/src/main/java/org/ethereum/sync/SyncManager.java
zilm13: Ahh, yeah, not the only thing that could be upgraded to streams, didn't try it. Big enough commit already )

PR #966, opened 2017-12-03
Increase coverage for .config.blockchain

weisheme:
- Replaced System.out with Logger in TransactionTest
- `TransactionTest.afterEIP158Test()` seems to expose a modeling issue. The required EIP-155 change of `v = CHAIN_ID * 2 + 36` is never stored in Transaction.signature field.


mkalinin:
@weisheme Nice! Much appreciate you keep working on tests! 
Regarding your comment on EIP-155 requirement. That `chainId` derivative is not assumed to be stored in `signature.v`. This feature is implemented in [getEncoded()](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java#L495-L496) method. And it may be worth to check.

Issue #963, opened 2017-12-01
Small configuration bugs found by tests


Nashatyrev:
See PR #946

PR #946, opened 2017-10-31
Increase test coverage for config

weisheme:
Not sure if a PR like this is appreciated at all, but at least for me writing tests is a great way to get to know the codebase. I started with `org.ethereum.config` and `org.ethereum.config.net`

I only made changes to test code. Production code is untouched.

Potentially found 3 small bugs related to configuration:
- NodeFilterTest's acceptNullIpPatternAsCatchAllForNodes and doNotAcceptNullIpPatternAsCatchAllForInetAddresses seem to indicate a discrepancy between InetAddress and Node acceptance. Smells like a potential bug.
- NodeFilterTest's acceptInvalidNodeHostnameWhenUsingWildcard and doNotAcceptInvalidNodeHostnameWhenUsingPattern seem to indicate a discrepancy in acceptance logic. Smells like a potential bug.
- SystemPropertiesTest's testExposeBugWhereNonHexEncodedIsAcceptedWithoutValidation seems to indicate that private key validation is not correct.

PR #967, opened 2017-12-03
#946 - Fixes for small configuration bugs found by tests

weisheme:
Could you please confirm that my reasoning is correct and that my bug fix is what you expected?

1. No longer accept nodes where the hostname cannot be resolved through `InetAddress.getByName()` **even** when
a `hostIpPattern = null` (wildcard) was provided.
1. Now accepts any InetAddress when `hostIpPattern = null` (wildcard) 
1. Only accept provided privateKey when it's Hex encoded

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/config/NodeFilter.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/NodeFilter.java
@@ -72,8 +72,12 @@ public class NodeFilter {
 
         public boolean accept(Node node) {
             try {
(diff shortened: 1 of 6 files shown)


mkalinin:
Should resolve https://github.com/ethereum/ethereumj/issues/963

mkalinin:
@Nashatyrev please, take a look

Nashatyrev:
Sounds reasonable

mkalinin:
@weisheme could you pls stick your branch with `develop`?

PR #969, opened 2017-12-09
Clean up Account class

tinchou:
1. Removed lots of unused code
2. Removed Spring DI annotations since this class was never injected as a component

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/Account.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Account.java
@@ -18,77 +18,19 @@
 package org.ethereum.core;
 
 import org.ethereum.crypto.ECKey;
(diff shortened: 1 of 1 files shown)


tinchou:
@mkalinin how does this look to you?

zilm13:
Hello

We have started detach of RPC from EthereumJ some time ago but have not finished it. I've created an issue #976 so we are going to take some action soon. As Action.class is used only in RPC implementation, whole class should be removed, but it is used in our own Ethereum Harmony and we should do it synchronously with Harmony update. For example, part of your commit would make develop version of Harmony broken. 

So, I ignore this commit but we will remove Account.class and everything related only to RPC from EthereumJ soon and it will be moved to [Ethereum Harmony](https://github.com/ether-camp/ethereum-harmony)

PR #940, opened 2017-10-11
fix #917 : fix bug in decodeBigInteger & refactor encodeElement

mlahia:

PR #975, opened 2017-12-14
Fixing some RLP bugs

zilm13:
Using pull requests #936, #940 
TrieImpl modifications in #936 are thrown away as it should be already fixed by fix constructor in https://github.com/ethereum/ethereumj/commit/08bef32
This bug is referenced in #907
Also issue #917 should be affected.
@mkalinin please, take a look.

Maybe, you have any additional test ideas.


ethereumj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -925,24 +978,30 @@ else if (srcData.length == 1 && (srcData[0] & 0xFF) < 0x80) {
...
+                data[i] = (byte) (tmpLength & 0xFF);
+                tmpLength = tmpLength >> 8;
             }
-            // first byte = F7 + bytes.length
-            byte[] data = Arrays.copyOf(srcData, srcData.length + 1 + byteNum);
-            System.arraycopy(data, 0, data, 1 + byteNum, srcData.length);
-            data[0] = (byte) (OFFSET_LONG_ITEM + byteNum);
-            System.arraycopy(lenBytes, 0, data, 1, lenBytes.length);
+
+            // at last copy the number bytes after its length
```
mkalinin: It may not only be a `number`

ethereumj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -1144,22 +1203,67 @@ else if (srcData.length == 1 && (srcData[0] & 0xFF) < 0x80) {
...
+
+            // [0x01, 0xb7] - 1-55 bytes item
+        } else if ((data[index] & 0xFF) <= OFFSET_LONG_ITEM) {
+
+            byte[] valueBytes = new byte[length];
+            System.arraycopy(data, index+1, valueBytes, 0, length);
+            return valueBytes;
+
+            // [0xb8, 0xbf] - 56+ bytes item,
+            // for 0xc0 and greater exception is thrown before in {@link #calculateItemLength}
```
mkalinin: Not sure that it is good practice to relay on exception thrown by another method. Somebody may not consider that fact and make changes that will break designed behavior.

ethereumj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -151,125 +150,153 @@ private static byte decodeOneByteItem(byte[] data, int index) {
...
-        System.arraycopy(data, index, valueBytes, 0, length);
-        return new BigInteger(1, valueBytes);
+        final byte[] valueBytes = decodeItemBytes(data, index);
+
+        if (valueBytes.length == 0) {
+            // shortcut
+            return BigInteger.ZERO;
+        } else {
+            BigInteger res = new BigInteger(1, valueBytes);
+            if (res.compareTo(BigInteger.ZERO) < 0) throw new RuntimeException("not a valid number");
```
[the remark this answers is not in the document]
zilm13: Ouch, nice catch )

ethereumj-core/src/main/java/org/ethereum/util/RLP.java
zilm13: Fixed

ethereumj-core/src/main/java/org/ethereum/util/RLP.java
zilm13: It's not about a type

PR #984, opened 2017-12-22
Feature events api

zilm13:
@mkalinin please take a look
Added events api + block replay
also removed obsolete networks settings


ethereumj-core/src/main/java/org/ethereum/listener/BlockReplay.java
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Replay blocks synchronously
+     */
+    public void replay() {
+        long lastBlock = blockStore.getMaxNumber();
+        logger.info("Replaying blocks from " + firstBlock + ", current best block: " + lastBlock);
+        int cnt = 0;
+        long num = firstBlock;
+        while(true) {
```
mkalinin: `while (!replayComplete)` is better if it does not break the logic (as I can see it does not)

ethereumj-core/src/main/java/org/ethereum/listener/BlockReplay.java
```diff
@@ -0,0 +1,221 @@
...
+                    logger.info("Replayed " + cnt + " blocks so far. Current block: " + num);
+                }
+            }
+
+            synchronized (this) {
+                if (onBlockBuffer.size() < onBlockBuffer.maxSize()) {
+                    replayComplete = true;
+                    break;
+                } else {
+                    long newLastBlock = blockStore.getMaxNumber() - 192;
```
mkalinin: is `192` a magic number? why exactly `192` but not the other one?

ethereumj-core/src/main/java/org/ethereum/listener/BlockReplay.java
zilm13: Sounds good

ethereumj-core/src/main/java/org/ethereum/listener/EventListener.java
```diff
@@ -0,0 +1,304 @@
...
+                    '}';
+        }
+    }
+
+    protected LogFilter contractAddressFilter;
+    protected CallTransaction.Contract contract;
+    protected PendingStateImpl pendingState;
+
+    // txHash => PendingEvent
+    protected Map<ByteArrayWrapper, PendingEvent> pendings = new LinkedHashMap<>();
```
mkalinin: Maybe wrap it by `ByteArrayMap` for convenience?

ethereumj-core/src/main/java/org/ethereum/listener/EventListener.java
```diff
@@ -0,0 +1,304 @@
...
+                            // with INCLUDED state
+                            // if Tx was included into fork block PendingState should call onPendingTransactionUpdateImpl
+                            // with PENDING state
+                            pendingState.trackTransaction(receipt.getTransaction());
+                        }
+                    }
+                }
+            }
+
+            if (blockSummary.getTotalDifficulty().compareTo(lastTotDiff) > 0) {
```
mkalinin: Isn't it a better idea to supply `BlockSummary` with `ImportStatus` to indicate whether imported block is best or not. This small feature might be useful in many other places and save us from keeping this redundant code and data structures

ethereumj-core/src/main/java/org/ethereum/listener/EventListener.java
```diff
@@ -0,0 +1,304 @@
...
+
+    // checks the Tx receipt for the contract LogEvents
+    // initiated on [onPendingTransactionUpdateImpl] callback only
+    private synchronized void onReceipt(TransactionReceipt receipt, Block block, PendingTransactionState state) {
+        ByteArrayWrapper txHashW = new ByteArrayWrapper(receipt.getTransaction().getHash());
+        if (contractAddressFilter.matchBloom(receipt.getBloomFilter())) {
+            int txCount = 0; // several transactions are possible withing a single Ethereum Tx
+            List<EventData> matchedTxs = new ArrayList<>();
+            for (LogInfo logInfo : receipt.getLogInfoList()) {
+                if (contractAddressFilter.matchBloom(logInfo.getBloom()) &&
```
mkalinin: Current implementation doesn't allow user to instantiate a listener which does topics filtering out of the box. I think it worth to be extended with that feature.

ethereumj-core/src/main/java/org/ethereum/listener/EventListener.java
zilm13: Cool, didn't know about it.

ethereumj-core/src/main/java/org/ethereum/listener/EventListener.java
zilm13: Good idea, but I'd rather put this logic in BlockSummary, because ImportState couldn't be always kept in BlockSummary, as summary may be null. I'll push this approach and we could discuss it, if it doesn't look ok.

ethereumj-core/src/main/java/org/ethereum/listener/EventListener.java
zilm13: Do you think we need such kind of low-level listener in addition to parsed events listener?

PR #986, opened 2017-12-25
Feature/db upgrade

mkalinin:


ethereumj-core/src/main/resources/ethereumj.conf
```diff
@@ -203,6 +203,7 @@ database {
         # controls how much last block states are not pruned
         # it is not recommneded to set this value below 192
         # as it can prevent rebranching from long fork chains
+        # consider enlarging cache.pruneWindowCacheSize if you want to increase maxDepth
```
zilm13: property is pruneCacheSize

ethereumj-core/src/main/resources/ethereumj.conf
```diff
@@ -234,10 +235,16 @@ cache {
...
+    # the size of read cache maintaining prune entries, in MBytes
+    # 8 MB of that cache should be enough to maintain a 256-blocks window
+    # window size is defined by database.prune.maxDepth
+    pruneCacheSize = 8
+
     # maximum size (in Mb) the state bloom fiter can grow up to
     # when reaching this threshold the bloom filter
     # is turned off forever
     # 128M can manage approx up to 50M of db entries
+    # deprecated, states source is no more backed with bloom filter
```
zilm13: may be capitalize DEPRECATED or smth like this, looks like a wall of text
same for the next "deprecated" note

ethereumj-core/src/main/java/org/ethereum/db/PeerSource.java
```diff
@@ -18,10 +18,7 @@
 package org.ethereum.db;
 
 import org.apache.commons.lang3.tuple.Pair;
-import org.ethereum.datasource.DataSourceArray;
-import org.ethereum.datasource.ObjectDataSource;
-import org.ethereum.datasource.Serializer;
-import org.ethereum.datasource.Source;
+import org.ethereum.datasource.*;
```
zilm13: imports: LevelDbDataSource unused
do we use * imports?

ethereumj-core/src/main/java/org/ethereum/db/RepositoryRoot.java
```diff
@@ -22,6 +22,7 @@
 import org.ethereum.datasource.*;
 import org.ethereum.trie.*;
 import org.ethereum.vm.DataWord;
+import org.spongycastle.util.encoders.Hex;
```
zilm13: unused import

ethereumj-core/src/test/java/org/ethereum/datasource/TrieNodeSourceIntegrationTest.java
```diff
@@ -0,0 +1,184 @@
...
+import static org.junit.Assert.assertNull;
+
+/**
+ * @author Mikhail Kalinin
+ * @since 05.12.2017
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(loader = AnnotationConfigContextLoader.class)
+@NoAutoscan
+@Ignore
```
zilm13: Why ignored? It's not very long

ethereumj-core/src/main/resources/ethereumj.conf
mkalinin: good idea 👍

ethereumj-core/src/main/java/org/ethereum/db/PeerSource.java
mkalinin: no, we don't

ethereumj-core/src/test/java/org/ethereum/datasource/TrieNodeSourceIntegrationTest.java
mkalinin: Decided to create similar integration test which uses memory instead of disk and don't need whole context to be involved

PR #989, opened 2018-01-15
Fix ignored test

weisheme:
Fixed an `@Ignore` on `TrieTest.testMassiveDeterministicUpdate`. The comparison of the resulting byte-array wasn't done correctly


mkalinin:
thanks!

PR #993, opened 2018-01-22
Bytecode compiler for tests

ajlopez:
A minor improvement for tests. An string with mnemonic opcodes could be compiled to a byte array of bytecodes.

As an example of use, I modified the VMTest.java. If needed, I could complete its modification in some test with long string of bytecodes (like codecopy and extcodecopy tests)

Maybe the bytecompiler could be used in other tests


ajlopez:
Sorry, I don't know if the failed travis is my fault or not, any suggestions?

zilm13:
@ajlopez It's not your fault, there are few json tests that could fail due to github api issues.
We will check your PR soon.

zilm13:
Wow, wonderful!
Could you also move 
ethereumj-core/src/test/java/org/ethereum/test/BytecodeCompiler.java
ethereumj-core/src/test/java/org/ethereum/test/BytecodeCompilerTest.java
to ethereumj-core/src/test/java/org/ethereum/vm . I think it will be better place for it as we don't need it for anything except vm and it's a little puzzlying location currently.

zilm13:
@ajlopez Thank you!

PR #998, opened 2018-01-23
Use ByteUtil to convert bytes to number

ajlopez:
In the current code, there are some lines where a byte array is converted to a number (integer, long, big integer), with an explicit checking of array null.

That functionality (null checking and convert) is already in ByteUtils.

This pull request change those lines to us e ByteUtil methods

The conversion to BigInteger was modified to consider array null, with an additional test to express expected behavior.


zilm13:
@ajlopez thank you! I will add a bit about null in ByteUtil comments to clarify it.

PR #999, opened 2018-01-25
optimize imports and reformat code

gabhi:
optimize imports and reformat code


zilm13:
Sorry, we are not accepting reformatting PRs, especially with so many files involved. It's impossible to check such big commits and there are no reasons to put production code at risk.

PR #1001, opened 2018-01-25
CLI Interface Refactor

ajlopez:
Some refactors, suggested by a linter (Sonar):

- Private default constructor, to avoid creation
- Constant string comparison, using constant.equals
- Extract methods to process arguments

I think it could be improved. For example, to have a class that processes an array of arguments, and returns a map of key values... and separate the apply of those key values to SystemProperties. In this way, we could have tests with expected behavior and better code coverage

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/cli/CLIInterface.java
+++ b/ethereumj-core/src/main/java/org/ethereum/cli/CLIInterface.java
@@ -17,6 +17,7 @@
  */
 package org.ethereum.cli;
 
+import org.apache.commons.lang3.BooleanUtils;
 import org.ethereum.config.SystemProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


zilm13:
@ajlopez Could you, please, merge latest develop into your branch and push it. We had an issues with tests, now it's fixed.

ethereumj-core/src/main/java/org/ethereum/cli/CLIInterface.java
```diff
@@ -99,10 +81,81 @@ public static void call(String[] args) {
...
+
+        Boolean resetFlag = interpret(reset);
+        logger.info("Resetting db set to [{}]", resetFlag);
+        cliOptions.put(SystemProperties.PROPERTY_DB_RESET, resetFlag.toString());
+
+        return true;
+    }
+
+    private static Boolean interpret(String arg) {
+        if ("on".equals(arg) || "true".equals(arg) || "yes".equals(arg)) return true;
```
zilm13: Could you replace both lines in favor of BooleanUtils.toBooleanObject(String) with throwing error when null returned. I guess it would be safer.

ajlopez: Yes! Done

zilm13: Add something like
```
if (resetFlag == null) {
    throw new Error("Can't interpret the answer: " + arg);
}
```
You've lost this case

ajlopez: Sure, good catch... done

PR #1002, opened 2018-01-25
Fixing Github tests with retries

zilm13:
Let's try it.
If it doesn't help, we could add sleep before retries.
UPDATE: Also it's better to rewrite loadJSONsFromCommit to perform retry only on failed files, not on whole request (if current solution does not help)


zilm13:
It runs smoothly but with 2 minutes overhead. Will improve a bit later.

zilm13:
I think it's ok now.
@mkalinin please, check it.

mkalinin:
looks good 👍

Issue #1006, opened 2018-02-07
Add support for option --allow-paths


tglaeser:
It seems that option `--allow-paths` is not supported by _ethereumj_.

As a result no Solidity contract importing from a parent or sibling directory can be compiled with `org.ethereum.solidity.compiler.SolidityCompiler`. E.g. trying to compile most contracts from [OpenZeppelin](https://github.com/OpenZeppelin/zeppelin-solidity/tree/master/contractsl) will fail to compile with an error like...
```
.../crowdsale/CappedCrowdsale.sol:3:1: Error: Source ".../math/SafeMath.sol" not found: File outside of allowed directories.
  import "../math/SafeMath.sol";
  ^----------------------------^
```
Any plans to address this? Thanks.

tglaeser:
In order to address this issue, I'm proposing the following...

1. Update dependency `org.ethereum:solcJ-all` to version `0.4.19`
2. Replace the fields from `enum` type `SolidityCompiler.Options` with new fields of the same name but different type implementing new `interface` `SolidityCompiler.Option`.

**Note** the naming above, Option**s** vs. Option.

This approach would ensure that for a consumer referencing `static` fields `org.ethereum.solidity.compiler.SolidityCompiler.Options.*` the change is non-breaking. For a consumer maintaining a list of available options however the type would need to be adjusted from `SolidityCompiler.Options` to `SolidityCompiler.Option`.

tglaeser:
Merging #1010 completes the issue.

PR #1010, opened 2018-02-16
Add support for option --allow-paths

tglaeser:
Following through on my proposal as outlined in ticket #1006.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/CallTransaction.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/CallTransaction.java
@@ -86,6 +86,13 @@ public class CallTransaction {
         }
     }
 
+    public enum StateMutabilityType {
+        pure,
+        view,
+        nonpayable,
+        payable
+    }
+
     public enum FunctionType {
         constructor,
         function,
@@ -101,6 +108,7 @@ public class CallTransaction {
         public Param[] inputs = new Param[0];
         public Param[] outputs = new Param[0];
         public FunctionType type;
+        public StateMutabilityType stateMutability;
 
         private Function() {}
 
--- a/ethereumj-core/src/main/java/org/ethereum/solidity/compiler/CompilationResult.java
+++ b/ethereumj-core/src/main/java/org/ethereum/solidity/compiler/CompilationResult.java
@@ -48,6 +48,9 @@ public class CompilationResult {
         }
     }
 
+    /**
+     * @return the contract's path given this compilation result contains exactly one contract
+     */
     @JsonIgnore public Path getContractPath() {
         if (contracts.size() > 1) {
             throw new UnsupportedOperationException("Source contains more than 1 contact. Please specify the contract name. Available keys (" + getContractKeys() + ").");
@@ -57,6 +60,9 @@ public class CompilationResult {
         }
     }
 
+    /**
+     * @return the contract's name given this compilation result contains exactly one contract
+     */
     @JsonIgnore public String getContractName() {
         if (contracts.size() > 1) {
             throw new UnsupportedOperationException("Source contains more than 1 contact. Please specify the contract name. Available keys (" + getContractKeys() + ").");
@@ -66,6 +72,10 @@ public class CompilationResult {
         }
     }
 
+    /**
+     * @param contractName The contract name
+     * @return the first contract found for a given contract name; use {@link #getContract(Path, String)} if this compilation result contains more than one contract with the same name
+     */
     @JsonIgnore public ContractMetadata getContract(String contractName) {
         if (contractName == null && contracts.size() == 1) {
             return contracts.values().iterator().next();
@@ -79,17 +89,28 @@ public class CompilationResult {
                 return entry.getValue();
             }
         }
-        throw new UnsupportedOperationException("Source contains more than 1 contact. Please specify a valid contract name. Available keys (" + getContractKeys() + ").");
+        throw new UnsupportedOperationException("No contract found with name '" + contractName + "'. Please specify a valid contract name. Available keys (" + getContractKeys() + ").");
     }
 
+    /**
+     * @param contractPath The contract path
+     * @param contractName The contract name
+     * @return the contract with key {@code contractPath:contractName} if it exists; {@code null} otherwise
+     */
     @JsonIgnore public ContractMetadata getContract(Path contractPath, String contractName) {
         return contracts.get(contractPath.toAbsolutePath().toString() + ':' + contractName);
     }
 
+    /**
+     * @return all contracts from this compilation result
+     */
     @JsonIgnore public List<ContractMetadata> getContracts() {
         return new ArrayList<>(contracts.values());
     }
 
+    /**
+     * @return all keys from this compilation result
+     */
     @JsonIgnore public List<String> getContractKeys() {
         return new ArrayList<>(contracts.keySet());
     }
--- a/ethereumj-core/src/main/java/org/ethereum/solidity/compiler/CompilationResult.java
+++ b/ethereumj-core/src/main/java/org/ethereum/solidity/compiler/CompilationResult.java
@@ -25,10 +25,10 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Collection;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class CompilationResult {
@@ -50,7 +50,7 @@ public class CompilationResult {
 
     @JsonIgnore public Path getContractPath() {
         if (contracts.size() > 1) {
-            throw new UnsupportedOperationException("Source contains more than 1 contact. Please specify the contract name. Available keys (" + getContractKeys().values() + ").");
+            throw new UnsupportedOperationException("Source contains more than 1 contact. Please specify the contract name. Available keys (" + getContractKeys() + ").");
         } else {
             String key = contracts.keySet().iterator().next();
             return Paths.get(key.substring(0, key.lastIndexOf(':')));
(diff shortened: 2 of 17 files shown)


tglaeser:
@mkalinin @zilm13 Any objections to get those changes merged into develop branch?

ethereumj-core/src/main/java/org/ethereum/core/CallTransaction.java
```diff
@@ -101,6 +108,7 @@ public String getType() {
         public Param[] inputs = new Param[0];
         public Param[] outputs = new Param[0];
         public FunctionType type;
+        public StateMutabilityType stateMutability;
```
zilm13: I cannot find any usage for it. Do we need it?

ethereumj-core/src/main/java/org/ethereum/solidity/compiler/SolidityCompiler.java
```diff
@@ -38,38 +47,96 @@ public SolidityCompiler(SystemProperties config) {
         solc = new Solc(config);
     }
 
-    public static Result compile(File sourceDirectory, boolean combinedJson, Options... options) throws IOException {
+    public static Result compile(File sourceDirectory, boolean combinedJson, Option... options) throws IOException {
         return getInstance().compileSrc(sourceDirectory, false, combinedJson, options);
     }
 
-    public enum Options {
+    public static final class Options {
```
zilm13: Could you, please, add comment here, that it's here for backward compatibility

ethereumj-core/src/main/java/org/ethereum/solidity/compiler/SolidityCompiler.java
```diff
@@ -38,38 +47,96 @@ public SolidityCompiler(SystemProperties config) {
...
+        public static final OutputOption METADATA = OutputOption.METADATA;
+        public static final OutputOption ASTJSON = OutputOption.ASTJSON;
+    }
+
+    public interface Option extends Serializable {
+        String getValue();
+        String getName();
+    }
+
+    public static class ListOption implements Option {
```
zilm13: This code currently designed only for `--allow-paths`, maybe make it just AllowPathsOption or at least make `AllowPathsOption extends ListOption` or any other shortcut, so user will not need to write `ListOption("allow-paths", `

ethereumj-core/src/test/java/org/ethereum/core/ImportLightTest.java
```diff
@@ -372,15 +372,15 @@ public void createContractFork() throws Exception {
                 "}";
 
         StandaloneBlockchain bc = new StandaloneBlockchain();
-        SolidityContract parent = bc.submitNewContract(contractSrc, "Parent");
+        SolidityContract parent = bc.submitNewContract(contractSrc, "<stdin>:Parent");
```
zilm13: I think it's a bad thing with all this `<stdin>`'s.
We are making interface to proxy user from solc and add some simplicity and programming-language interface to interact with. Also all method declarations call it "Contract name" already. Plus we want it to work with several versions of solc.
I'd suggest modify CompilationResult.parse to remove "<stdin>:" part when it exists in the beginning and revert all method calls in tests. What do you think?

ethereumj-core/src/main/java/org/ethereum/solidity/compiler/SolidityCompiler.java
tglaeser: Will do.

ethereumj-core/src/main/java/org/ethereum/solidity/compiler/SolidityCompiler.java
tglaeser: You are absolutely right; I'll do the necessary changes.

ethereumj-core/src/test/java/org/ethereum/core/ImportLightTest.java
tglaeser: Well, I didn't spent a lot of thoughts on this, simply adjusted the tests to work with the new Solidity compiler version.

To me it seems that the creators of Solidity, with version [0.4.9](https://github.com/ethereum/solidity/releases/tag/v0.4.9), added some form of a poor men's namespace support; that is, requiring to prefix each contract with `${filename}:`.  Only if the contract is read from standard input instead, the prefix becomes `<stdin>:`. Please follow the link above for the release notes.

I don't see how the new contract's namespace information could be ignored in the production code. Thoughts/concerns?

zilm13: Ok, what I suggest:
1) Adding public String path into `ContractMetadata`. If there is anything on first line before `:`, it's saved there
2) `public Map<String, ContractMetadata> contracts` becomes `public Map<String, List<ContractMetadata>> contracts`

What do you think? 
We are braking backward compatibility but I see only weird ways to preserve it.
Anyway remembering path which is specific to filesystem to get contract from compilation is not a good thing, I think, and we need to get it over.

tglaeser: I understand your suggestion, but I believe that would not be as designed by the creators of Solidity.

To me it seems, since Solidity version `0.4.9`, that the compiler returns a contract key, not contract name anymore, which will **always** be prefixed with `${filename}:` or `<stdin>:`, that's a breaking change at the side of the Solidity compiler. You **cannot** ditch this information and use the contract name instead of the contract key as the key in `Map<String, ContractMetadata>`. This breaking change was introduced to allow contracts of the same name but different location to coexists, e.g. map entries with keys `/foo/file1.sol:test`, `/foo/bar/file2.sol:test`.

We need to agree on this first before we can think about how to better address this change.

zilm13: Look, now contract name just contains more information: path (or `<stdin>`) and contract name. I just suggest to store it separately in our Java model. Contract name is not {filename}:realContractName. It's just new part of information, they've added in output in such way. You are not able to call one contract from another with such prefix, so it's not a part of contract name, yeah?

tglaeser: What do you want to use as key in `Map<String, ContractMetadata>`, `${filename}:realContractName` or `realContractName`?

zilm13: I suggest modify it to `Map<String, List<ContractMetadata>>`
usually it will be list with 1 `ContractMetadata`, but when there are two or more, user will get both in list and differentiate them by public field `path` which we'll add to  `ContractMetadata`. 
Or even better `Map<String, Map<String, ContractMetadata>>` -> `ContractName: Path: ContractMetadata`. As `path` is not a part of ContractMetadata by meaning.

zilm13: `Map<String, LinkedMap<String, ContractMetadata>>` so it could be accessible in simplest way: `contracts.get("MyContract").getValue(0)` 
@mkalinin what do you think?

tglaeser: I can see how `Map<String, Map<String, ContractMetadata>>` could possibly work, but at what price? That seems to be a massive incompatibility for no good reason. The Solidity compiler returns a contract key on purpose but we then use the contract name instead as the `key` in our `Map<String, ?>`, no matter of the `value` type.
We clearly have opposite opinions on this topic, you prefer to use the contract name as `key` so that the tests continue to work while I believe a `key` is a `key` is a unique `key`. Any chance that you can bring in someone else from the team to be the tiebreaker?

zilm13: Yeah, I've called @mkalinin for his opinion, he will reply later.
I don't like paths (or combined key as originally comes from solc) as a first key because paths are even different on different filesystems and we make life of all users more difficult for rare cases when someone needs to compile the same names at once. Let's see what Michael will say.

tglaeser: Yeah, I noticed, our messages somehow overlapped. I can relate to the discontent towards `java.io.File`, so maybe `Map<URI, ?>` would actually better reflect reality.

tglaeser: One more thing, doing `contracts.get("MyContract").getValue(0)` as suggested by @zilm13 above would allow keeping the current tests more or less the same, however any general provider of this functionality would still have to check the entries in `LinkedMap<String, ContractMetadata>` in order to make sure that he doesn't miss any contract. It just unnecessarily complicates the code IMHO.

mkalinin: Both of you have thoughts meaningful to me. I agree with @zilm13 that contract name is just a name and it's incorrect to prefix it with contract source, cause that will definitely confuse the user. On the other hand @tglaeser is right pointing that `CompilationResult` contains solidity output where `key` in `contracts` map is not a contract name but its identifier. So, what would I do: hide `CompilationResult.contracts` and provide getters to access it. I.e. something like `getContract(String name)` for general usage which returns first matching contract with no matter for path, `getContract(String name, String path)` - this method requires additional work in path transformation depending on the OS (need to check how solc handles that). Also, this change requires additional getter which returns just the first contract from that map (I saw that this case is also can take place). That is my opinion guys and that solution combines both of your suggestions, hence it should work :)

tglaeser: Thanks for the input. I'll go forward and make the necessary changes and update the pull request...

tglaeser:
@zilm13 @mkalinin Please let me know in case you have additional thoughts/suggestions...

ethereumj-core/src/main/java/org/ethereum/solidity/compiler/CompilationResult.java
```diff
@@ -42,6 +48,52 @@ public static CompilationResult parse(String rawJson) throws IOException {
...
+    @JsonIgnore public String getContractName() {
+        if (contracts.size() > 1) {
+            throw new UnsupportedOperationException("Source contains more than 1 contact. Please specify the contract name. Available keys (" + getContractKeys() + ").");
+        } else {
+            String key = contracts.keySet().iterator().next();
+            return key.substring(key.lastIndexOf(':') + 1);
+        }
+    }
+
+    @JsonIgnore public ContractMetadata getContract(String contractName) {
```
zilm13: Could you add javadoc to getters, especially to this one. There should be mentioned, that if there are more than 1 contract with the same name, only first will be returned, so you should use `{@link getContract(Path contractPath, String contractName)}` for this cases.

ethereumj-core/src/test/java/org/ethereum/solidity/CompilerTest.java
```diff
@@ -86,27 +88,87 @@ public void defaultFuncTest() throws IOException {
...
+        SolidityCompiler.Option allowPathsOption = new SolidityCompiler.Options.AllowPaths(Collections.singletonList(source.getParent().getParent().toFile()));
+        SolidityCompiler.Result res = SolidityCompiler.compile(source.toFile(), true, ABI, BIN, INTERFACE, METADATA, allowPathsOption);
+        System.out.println("Out: '" + res.output + "'");
+        System.out.println("Err: '" + res.errors + "'");
+        CompilationResult result = CompilationResult.parse(res.output);
+
+        CompilationResult.ContractMetadata a = result.getContract(source, "test3");
+        CallTransaction.Contract contract = new CallTransaction.Contract(a.abi);
+        System.out.print(contract.functions[0].toString());
+    }
```
zilm13: could you assert here that there are 2 contracts totally and second is not null two and could be queried by name.

tglaeser:
@zilm13 Made the requested changes.

zilm13:
@tglaeser Thank you!

mkalinin:
guys, have you an idea of what went wrong? https://github.com/ether-camp/ethereum-harmony/issues/63

zilm13:
@mkalinin I'll fix it

PR #1015, opened 2018-02-22
Casper FFG test net compatibility

zilm13:
NO MERGE
A lot of refactoring to go.
Only for develop diff.


zilm13:
TODO: add Test for CasperTransactionExecutor with vote to check that gas is not used and rewards are correct

ethereumj-core/src/main/java/org/ethereum/casper/config/CasperBeanConfig.java
```diff
@@ -0,0 +1,102 @@
...
+import org.ethereum.manager.WorldManager;
+import org.ethereum.mine.BlockMiner;
+import org.ethereum.vm.program.invoke.ProgramInvokeFactory;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+@Configuration
+@EnableTransactionManagement
+public class CasperBeanConfig extends CommonConfig {
```
mkalinin: Will `@Import(DefaultConfig.class)` work here instead of inheritance? If yes then it might also keep us from adding `CasperConfig`.

ethereumj-core/src/main/java/org/ethereum/casper/config/CasperBeanConfig.java
```diff
@@ -0,0 +1,102 @@
...
+        return new CasperBlockchain(systemProperties());
+    }
+
+    @Override
+    @Bean
+    public TransactionExecutorFactory transactionExecutorFactory() {
+        return new CasperTransactionExecutorFactory();
+    }
+
+    class CasperTransactionExecutorFactory implements TransactionExecutorFactory {
```
mkalinin: Better keep an implementation in the same package where interface is placed in

ethereumj-core/src/main/java/org/ethereum/casper/core/CasperFacade.java
```diff
@@ -0,0 +1,96 @@
...
+        if (casper == null) {
+            contractAddress = Hex.toHexString(systemProperties.getCasperAddress());
+            String casperAbi = systemProperties.getCasperAbi();
+            casper = new CallTransaction.Contract(casperAbi);
+        }
+    }
+
+    public Object[] constCall(String func, Object... funcArgs) {
+        init();
+        ProgramResult r = ethereum.callConstantFunction(contractAddress,
```
mkalinin: Looks like a call to a function needs its own code abstraction marked by `@Scope("prototype")` for easy dependency resolving. Thus, client won't always need an `Ethereum` instance to call a function.

ethereumj-core/src/main/java/org/ethereum/casper/core/CasperTransactionExecutor.java
```diff
@@ -0,0 +1,145 @@
...
+        } else {
+            call();
+        }
+    }
+
+    private boolean isCasperVote() {
+        return isCasperVote(tx, ((CasperProperties) config).getCasperAddress());
+    }
+
+    public static boolean isCasperVote(Transaction transaction, byte[] casperAddress) {
```
mkalinin: I think `CasperFacade` better fits this logic

ethereumj-core/src/main/java/org/ethereum/casper/core/CasperTransactionExecutor.java
```diff
@@ -0,0 +1,145 @@
...
+        if (!Arrays.equals(transaction.getSender(), Transaction.NULL_SENDER))
+            return false;
+        if (casperAddress == null)
+            return false;
+        if (!Arrays.equals(transaction.getReceiveAddress(), casperAddress))
+            return false;
+
+        byte[] dataCopy = new byte[4];
+        System.arraycopy(transaction.getData(), 0, dataCopy, 0, 4);
+        return Arrays.equals(dataCopy, new byte[] {(byte) 0xe9, (byte) 0xdc, 0x06, 0x14});
```
mkalinin: Use `FastBytesComparison` to avoid `dataCopy`. Also, let's keep constant array in a static class member.

ethereumj-core/src/main/java/org/ethereum/casper/core/genesis/CasperStateInit.java
```diff
@@ -0,0 +1,200 @@
...
+        blockchain.setTotalDifficulty(genesis.getCumulativeDifficulty());
+
+        logger.info("Genesis block loaded");
+    }
+
+    /**
+     * @return key: Casper contract address
+     *       value: Casper state initialization transactions
+     */
+    public Pair<byte[], List<Transaction>> makeInitTxes() {
```
mkalinin: I think these Txes is better to keep in `CasperFacade` responsibility scope

ethereumj-core/src/main/java/org/ethereum/casper/core/genesis/CasperStateInit.java
```diff
@@ -0,0 +1,200 @@
...
+
+        return new Pair<>(casperAddress, txs);
+    }
+
+    private void init() {
+        Repository repo = repository.getSnapshotTo(null);
+        Genesis.populateRepository(repo, genesis);
+
+        // Metropolis dummy
+        repo.saveCode(Hex.decode("0000000000000000000000000000000000000010"), Hex.decode("6000355460205260206020f3"));
```
mkalinin: It worth to add a `contracts` section in Genesis file format to keep such specific things there.

ethereumj-core/src/main/java/org/ethereum/casper/manager/CasperWorldManager.java
```diff
@@ -0,0 +1,50 @@
...
+    public CasperWorldManager(SystemProperties config, Repository repository, Blockchain blockchain) {
+        super(config, repository, blockchain);
+    }
+
+    @PostConstruct
+    @Override
+    protected void init() {
+        super.init();
+        casper.setEthereum(ethereum);
+        if (Boolean.TRUE.equals(((CasperProperties) config).getCasperValidatorEnabled())) {
```
mkalinin: Just add `@Autowired public void setConfig(SystemProperties config) { ... }` method to 
`CasperValidatorService` like it's done in other similar places

ethereumj-core/src/main/java/org/ethereum/casper/mine/CasperBlockMiner.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class CasperBlockMiner extends BlockMiner {
+
+    public CasperBlockMiner(SystemProperties config, CompositeEthereumListener listener,
+                            Blockchain blockchain, PendingState pendingState) {
+        super(config, listener, blockchain, pendingState);
+    }
+
+    @Override
+    protected boolean isAcceptableTx(Transaction tx) {
```
mkalinin: It turns that we need Tx validator sets managed by Spring config and instantiated there for each specific case. This would make `BlockMiner` and `PendingStateImpl` clearer

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
```diff
@@ -867,7 +876,8 @@ private BlockSummary applyBlock(Repository track, Block block) {
             stateLogger.debug("apply block: [{}] tx: [{}] ", block.getNumber(), i);
 
             Repository txTrack = track.startTracking();
-            TransactionExecutor executor = new TransactionExecutor(tx, block.getCoinbase(),
+            TransactionExecutorFactory txFactory = commonConfig.transactionExecutorFactory();
```
mkalinin: Inject this factory with Spring?

ethereumj-core/src/main/java/org/ethereum/core/PendingStateImpl.java
```diff
@@ -183,6 +196,10 @@ public void addPendingTransaction(Transaction tx) {
         return newPending;
     }
 
+    protected boolean receiptIsValid(TransactionReceipt receipt) {
```
mkalinin: The same concern as for Tx validators.

ethereumj-core/src/main/java/org/ethereum/db/BlockStore.java
```diff
@@ -64,6 +64,11 @@
 
     void reBranch(Block forkBlock);
 
+    // FIXME: Do something universal instead like BlockInfoV2
+    void saveMeta(String name, byte[] value);
```
mkalinin: Why not a separate checkpoint storage?

ethereumj-core/src/main/java/org/ethereum/db/RepositoryRoot.java
```diff
@@ -129,7 +129,9 @@ public synchronized void flush() {
 
     @Override
     public Repository getSnapshotTo(byte[] root) {
-        return new RepositoryRoot(stateDS, root);
+        RepositoryRoot repo = new RepositoryRoot(stateDS, root);
+        repo.config = this.config;
```
mkalinin: pls, move this argument to the constructor

ethereumj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
```diff
@@ -247,7 +278,8 @@ public BlockSummary replayBlock(Block block) {
             for (Transaction tx : block.getTransactionsList()) {
 
                 Repository txTrack = track.startTracking();
-                org.ethereum.core.TransactionExecutor executor = new org.ethereum.core.TransactionExecutor(
+                TransactionExecutorFactory txFactory = commonConfig.transactionExecutorFactory();
```
mkalinin: Inject it as a regular Spring bean

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth62.java
```diff
@@ -195,8 +194,8 @@ public synchronized void sendStatus() {
             totalDifficulty = blockchain.getTotalDifficulty();
         }
 
-        StatusMessage msg = new StatusMessage(protocolVersion, networkId,
-                ByteUtil.bigIntegerToBytes(totalDifficulty), bestHash, config.getGenesis().getHash());
+        StatusMessage msg = new StatusMessage(protocolVersion, networkId,  // Casper genesis from config is not final
```
mkalinin: Maybe this change would gone if contracts' code was a part of genesis file

ethereumj-core/src/main/resources/ethereumj.conf
```diff
@@ -428,3 +428,11 @@ crypto {
 	hash.alg256="ETH-KECCAK-256"
 	hash.alg512="ETH-KECCAK-512"
 }
+
+consensus {
```
mkalinin: This definition is redundant I guess

ethereumj-core/src/main/java/org/ethereum/casper/config/CasperBeanConfig.java
zilm13: Not sure, will check. Expecting two beans of the same type by intuition.

ethereumj-core/src/main/java/org/ethereum/casper/config/CasperBeanConfig.java
zilm13: Why? We have implementation for different purpose. Interface is universal and used as interface.

ethereumj-core/src/main/java/org/ethereum/casper/core/CasperFacade.java
zilm13: Nice catch, will try to implement.

ethereumj-core/src/main/java/org/ethereum/casper/core/CasperTransactionExecutor.java
zilm13: Agree, got this idea yesterday but didn't implement yet.

ethereumj-core/src/main/java/org/ethereum/casper/core/CasperTransactionExecutor.java
zilm13: Ok

ethereumj-core/src/main/java/org/ethereum/casper/core/genesis/CasperStateInit.java
zilm13: Originally it was a part of METROPOLIS, I could put it in genesis but not sure that it's the right place for it. Anyway it's a hack so any place is not right :)

ethereumj-core/src/main/java/org/ethereum/casper/manager/CasperWorldManager.java
zilm13: We'll fail if this action starts before postconstruct, during Autowired process, because not all dependencies are in place.

ethereumj-core/src/main/java/org/ethereum/casper/mine/CasperBlockMiner.java
zilm13: Yeah like BlockHeaderValidator etc. Had such idea, agree.

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
zilm13: It'll be a bit simpler in using, but increases difficulty for testing. I will try to find good trade-off in such cases.

ethereumj-core/src/main/java/org/ethereum/db/BlockStore.java
zilm13: Yeah, need to create it.

ethereumj-core/src/main/java/org/ethereum/db/RepositoryRoot.java
zilm13: Kk

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/Eth62.java
zilm13: Yeah, it should gone.

ethereumj-core/src/main/resources/ethereumj.conf
zilm13: forget about it. Will clean up.

ethereumj-core/src/main/java/org/ethereum/casper/core/genesis/CasperStateInit.java
zilm13: Yeah, looks like we don't need StateInit abstraction at all, will try to get rid of it.

ethereumj-core/src/main/java/org/ethereum/casper/manager/CasperWorldManager.java
mkalinin: ok, got it. then maybe it's better to wrap this check in `CasperValidatorService#init()` and just call `init()` here?

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
mkalinin: ok, maybe we will just get rid of that factory

ethereumj-core/src/main/java/org/ethereum/casper/manager/CasperWorldManager.java
zilm13: Ok, looks reasonable

zilm13:
FIXME: When we have tx in epoch switch block, its stored with wrong number in transactionStore, as there was tx with epoch switch before (which is not included in block txs)

zilm13:
TODO: 
1) should epoch switch costs gas? 
2) for failed casper vote, we are substituting fee from tx.sender, but the sender is fffffffffff.. ?

1) no
2) if a casper vote's sender is ffffffffff, and the casper vote fails, that's an invalid transaction, and if that tx is part of the block then the block is invalid. pyethapp is currently incorrectly implemented

zilm13:
1) postpone as we will need to restart testnet for it

zilm13:
+ add optional slashing condition watcher

mkalinin:
Closing it

PR #1017, opened 2018-02-27
Fix ClassCastException

feiqihang:
When the "peer.p2p.pingInterval" parameter is configured in the ethereumj.conf file, the exception "java.lang.ClassCastException: java.lang.Integer can not be cast to java.lang.Long" appears. After analyzing the code, we found that in the getAnyRef method of the com.typesafe.config.Config class, the java.lang.Integer type is returned when getting the numeric parameter. And, I think the unit for the pingInterval parameter is seconds, using the Integer type is appropriate.


mkalinin:
👍

PR #1043, opened 2018-03-29
upgrade guava

adridadou:
This change is [Reviewable](https://reviewable.io/reviews/ethereum/ethereumj/1043)


mkalinin:
Looks good 👍  One thing a little bit confusing me is the number of commits

mkalinin:
@adridadou can you please decouple guava update commits? It may be easier to create another PR for that.

PR #1052, opened 2018-04-15
No need to give arguments to println

kishansagathiya:
if all you want is to leave an empty line


kishansagathiya:
Just trying to get started with ethereumj with small fixes

PR #1056, opened 2018-04-16
Updated gas price tracker added

zilm13:
This is an updated version that returns more accurate data 
 in networks with large number of transactions like Ethereum MainNet.
 However it needs more CPU and memory resources for processing so it's not enabled by default.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
+++ b/ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
@@ -40,7 +40,7 @@ import java.util.List;
  */
 public class RecommendedGasPriceTracker extends EthereumListenerAdapter {
 
-    private static final long DEFAULT_PRICE = 1_000_000_000L;
+    private static final Long DEFAULT_PRICE = null;
     private static final int MIN_BLOCKS = 128;
     private static final int BLOCKS_RECOUNT = 1;
     private static final int MIN_TRANSACTIONS = 512;
@@ -50,7 +50,7 @@ public class RecommendedGasPriceTracker extends EthereumListenerAdapter {
     private CircularFifoQueue<Long> blockGasPricesExtra = new CircularFifoQueue<>(getMinTransactions());
 
     private int idx = 0;
-    private long recommendedGasPrice = getDefaultPrice();
+    private Long recommendedGasPrice = getDefaultPrice();
 
     public RecommendedGasPriceTracker() {
         blockGasPrices = new CircularFifoQueue<>(getMinBlocks());
@@ -115,7 +115,7 @@ public class RecommendedGasPriceTracker extends EthereumListenerAdapter {
      * Until this {@link #getDefaultPrice()} is returned
      * @return recommended gas price for transaction
      */
-    public long getRecommendedGasPrice() {
+    public Long getRecommendedGasPrice() {
         return recommendedGasPrice;
     }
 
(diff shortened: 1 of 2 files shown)


ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final int MIN_TRANSACTIONS = 512;
+    private static final int PERCENTILE_SHARE = 4;
+
+    private CircularFifoQueue<long[]> blockGasPrices;
+
+    private int idx = 0;
+    private Long recommendedGasPrice = getDefaultPrice();
+
+    public RecommendedGasPriceTracker() {
+        blockGasPrices = new CircularFifoQueue<>(Math.max(getMinTransactions(), getMinBlocks()));
```
mkalinin: Since `gasPrices` collection became not responsible for purging old Txes `CircularFifoQueue` starts to be redundant. Maybe it's better to use `LinkedList` here?

ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Override
+    public void onBlock(BlockSummary blockSummary) {
+        onBlock(blockSummary.getBlock());
+    }
+
+    private void onBlock(Block block) {
+        if (onTransactions(block.getTransactionsList())) {
+            ++idx;
+            if (idx == getBlocksRecount()) {
```
mkalinin: I'd rather use less strict `>=` condition here, just in case. Or another way is to keep using modulo operand if `idx` param would never reset

ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private synchronized boolean onTransactions(List<Transaction> txs) {
+        if (txs.isEmpty()) return false;
+
+        long[] gasPrices = new long[txs.size()];
+        for (int i = 0; i < txs.size(); ++i) {
+            gasPrices[i] = ByteUtil.byteArrayToLong(txs.get(i).getGasPrice());
+        }
+
+        while (blockGasPrices.size() >= getMinBlocks() &&
```
mkalinin: If `gasPrices` would be added before the loop then we got rid of extra `+ gasPrices.length` argument in the loop condition. But this would only work if intrusive `CircularFifoQueue` was replaced

mkalinin: If above changes are applied `>=` can be replaced with  `>` which should be more precise condition

ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
```diff
@@ -0,0 +1,179 @@
...
+        return blockGasPrices.stream().map(Array::getLength).mapToInt(Integer::intValue).sum();
+    }
+
+    private synchronized Long getGasPrice() {
+        int size = calcGasPricesSize();
+        // Don't override default value until we have minTransactions and minBlocks
+        if (size < getMinTransactions() ||
+                blockGasPrices.size() < getMinBlocks()) return null;
+
+        long[] difficulties = new long[size > getMinTransactions() ? size : getMinTransactions()];
```
mkalinin: Ternary operator used in array size definition can be omitted here

zilm13:
@mkalinin sounds reasonable. updated.

PR #1061, opened 2018-04-27
Fix slow import

mkalinin:


ethereumj-core/src/main/resources/ethereumj.conf
```diff
@@ -218,6 +218,11 @@ database {
         # 1_000_000: 5658 Mb
         maxDepth = 192
     }
+
+    # defines a number of opened files by db instance
+    # this number has significant impact on read amplification
+    # on the other hand it can force exceeding of user's limit, OS usually set it to 1024
```
zilm13: > OS usually set it to 1024

\+ "for all applications

ethereumj-core/src/main/java/org/ethereum/datasource/DbSettings.java
```diff
@@ -0,0 +1,33 @@
...
+ * Defines configurable database settings
+ *
+ * @author Mikhail Kalinin
+ * @since 26.04.2018
+ */
+public class DbSettings {
+
+    public static final DbSettings DEFAULT = new DbSettings();
+
+    int maxOpenFiles = 32;
```
zilm13: Maybe add 2 more vars as statics above and assign here from them?
Don't like a way magic constants are passed.

mkalinin: Agree. Don't like them either :)

mkalinin:
@zilm13 ready to be merged after the review

zilm13:
@mkalinin to be clear, LevelDB has `options.maxOpenFiles()`. Everything else looks cool!

PR #1062, opened 2018-05-02
More memory fixes

zilm13:
Not ready yet, don't merge


zilm13:
@mkalinin I think we could merge it after review
thread locks are not regular, I will try to catch them for a few days later.

PR #1063, opened 2018-05-02
Use isEmpty() to check for empty strings

kishansagathiya:
and Remove unnecessary StringUtils library

@mkalinin @zilm13 One more small change :)


zilm13:
@kishansagathiya `StringUtils.isEmpty` is null safe. I'd better keep it.

kishansagathiya:
@zilm13 Made changes as you asked for

Issue #1032, opened 2018-03-13
Replace `Hex.toHexString` in logging, toString etc to safer alternative


zilm13:
Fails on null, pain.

kishansagathiya:
@zilm13 This looks like something simple that I can pick up, is it?

zilm13:
@kishansagathiya sure, check `ByteUtil.toHexString`, it exists, but not used in all hex logging.

kishansagathiya:
Tried few different libraries, failed to find any null safe hex encoder.
 
@zilm13 
org.apache.commons.codec.binary.Hex.encodeHexString()
org.apache.soap.encoding.Hex.encode(null)
I guess spongycastle uses bouncycastle internally, but tried that anyway
org.bouncycastle.util.encoders.Hex.encode(null)

How about adding a null check before every call or create our own method that wraps this function with a null check, 
I believe if null, we would be throwing illigalArgumentError right?

zilm13:
@kishansagathiya what you don't like in `org.ethereum.util.ByteUtil.toHexString`which I mentioned before?

kishansagathiya:
@zilm13 I thought that is one more thing that is not null safe method :), 

Anyway org.ethereum.util.ByteUtil.toHexString is perfect. I will use it.

kishansagathiya:
@zilm13 
I believe you don't want to replace all Hex.toHexString() to ByteUtil.toHexUtil, right? Otherwise I can just do a replace all.

For something like https://github.com/ethereum/ethereumj/blob/4cdc799a2e957110656a38bf9a40769b5e9c2630/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java#L687
You would rather want to through an error instead of setting to null,
Right?

zilm13:
@kishansagathiya I want to change `Hex.toHexString()` to `ByteUtil.toHexUtil` in logging only. So we could see at least part of data in logs if smth is null (instead of Exception).

zilm13:
resolved with #1069

PR #1069, opened 2018-05-09
Issue #1032 Use null safe method for bytes to hex string conversion

kishansagathiya:
This is does not include all such cases
Fix for #1032 

@zilm13 Since I would be using the same approach throughout all such cases, i just wanted to get this validated if this approach looks good to you.


kishansagathiya:
@zilm13 If you are ok with static imports, I will use the same thing throughout all files wherever I make changes

kishansagathiya:
To avoid making this a huge PR, I have added changes only to the implementation and once this is merged I will make another for test, Is that cool?

Also, I have made changes only at those place where we are printing or logging some message ( i.e, logs, some println, exception or error messages etc ). Have avoided changing toString() functions. I can do that with subsequent PRs with making sure that my new change don't avoid throwing an error that should been thrown otherwise.

zilm13:
@kishansagathiya 
> Have avoided changing toString() functions

We need to change them too. It's object logging that fails in case of one field is null.

You'd better move file by file, so I could check that it's done for this file but as you already did it for many files, just add it to toString methods too and I will check everything together. From first view you did it ok, but it needs more attention with branch checkout before merge, so it's better to finish it for toString methods too or any other methods that are called from logging.

kishansagathiya:
@zilm13 ready for review. I hope none of the toString() methods are being used for anything except logging otherwise this might break things.

ethereumj-core/src/main/java/org/ethereum/datasource/rocksdb/RocksDbDataSource.java
```diff
@@ -40,6 +40,7 @@
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import static java.lang.System.arraycopy;
+import static org.ethereum.util.ByteUtil.toHexString;
```
zilm13: unused `org.spongycastle.util.encoders.Hex` import above

ethereumj-core/src/main/java/org/ethereum/net/eth/message/GetBlockBodiesMessage.java
```diff
@@ -25,6 +25,8 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import static org.ethereum.util.ByteUtil.toHexString;
```
zilm13: unused `org.spongycastle.util.encoders.Hex` import above

ethereumj-core/src/main/java/org/ethereum/net/eth/message/NewBlockMessage.java
```diff
@@ -25,6 +25,8 @@
 
 import java.math.BigInteger;
 
+import static org.ethereum.util.ByteUtil.toHexString;
```
zilm13: unused `org.spongycastle.util.encoders.Hex` import above

ethereumj-core/src/main/java/org/ethereum/net/eth/message/StatusMessage.java
```diff
@@ -25,6 +25,8 @@
 
 import java.math.BigInteger;
 
+import static org.ethereum.util.ByteUtil.toHexString;
```
zilm13: unused `org.spongycastle.util.encoders.Hex` import above

ethereumj-core/src/main/java/org/ethereum/net/rlpx/PongMessage.java
```diff
@@ -101,7 +97,7 @@ public String toString() {
         long currTime = System.currentTimeMillis() / 1000;
 
         String out = String.format("[PongMessage] \n token: %s \n expires in %d seconds \n %s\n",
-                Hex.toHexString(token), (expires - currTime), super.toString());
+        ByteUtil.toHexString(token), (expires - currTime), super.toString());
```
zilm13: correct spacing, please

ethereumj-core/src/main/java/org/ethereum/vm/program/invoke/ProgramInvokeFactoryImpl.java
```diff
@@ -179,19 +178,19 @@ public ProgramInvoke createProgramInvoke(Program program, DataWord toAddress, Da
...
+                    ByteUtil.toHexString(caller.getLast20Bytes()),
                     balance.toString(),
                     gasPrice.longValue(),
                     gas.longValue(),
-                    Hex.toHexString(callValue.getNoLeadZeroesData()),
-                    data == null ? "" : Hex.toHexString(data),
-                    Hex.toHexString(lastHash.getData()),
-                    Hex.toHexString(coinbase.getLast20Bytes()),
+                    ByteUtil.toHexString(callValue.getNoLeadZeroesData()),
+                    data == null ? "" : ByteUtil.toHexString(data),
```
zilm13: As you've changed it to null-safe, condition checking is no more required

ethereumj-core/src/main/java/org/ethereum/net/swarm/Manifest.java
```diff
@@ -1,28 +1,27 @@
...
+ * (at your option) any later version.
+ *
+ * The ethereumJ library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with the ethereumJ library. If not, see <http://www.gnu.org/licenses/>.
+ */
```
kishansagathiya: This is weird. The comment is not changed at all, but GitHub is showing it to be changed.

ethereumj-core/src/main/java/org/ethereum/samples/EventListenerSample.java
```diff
@@ -225,7 +225,7 @@ public void onPendingTransactionUpdate(TransactionReceipt txReceipt, PendingTran
     }
 
     public void requestFreeEther(byte[] addressBytes) {
-        String address = "0x" + Hex.toHexString(addressBytes);
+        String address = "0x" + toHexString(addressBytes);
```
kishansagathiya: `address` is used only for logging, so no harm changing it here.

ethereumj-core/src/main/java/org/ethereum/vm/program/InternalTransaction.java
```diff
@@ -17,11 +17,6 @@
  */
 package org.ethereum.vm.program;
 
-import static org.apache.commons.lang3.ArrayUtils.getLength;
-import static org.apache.commons.lang3.ArrayUtils.isEmpty;
-import static org.apache.commons.lang3.ArrayUtils.nullToEmpty;
-import static org.ethereum.util.ByteUtil.toHexString;
-
```
kishansagathiya: because duplicate imports

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1255,7 +1255,7 @@ else if (oldValue != null && newValue.isZero()) {
                     program.setHReturn(hReturn);
 
                     if (logger.isInfoEnabled())
-                        hint = "data: " + Hex.toHexString(hReturn)
+                        hint = "data: " + toHexString(hReturn)
```
kishansagathiya: `hint` is used only in logs

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1401,7 +1401,7 @@ private void dumpLine(OpCode op, long gasBefore, long gasCost, long memWords, Pr
             }
 
             int level = program.getCallDeep();
-            String contract = Hex.toHexString(program.getOwnerAddress().getLast20Bytes());
+            String contract = toHexString(program.getOwnerAddress().getLast20Bytes());
```
kishansagathiya: `contract` is used only in logs

zilm13:
Thank you!

PR #1072, opened 2018-05-10
Replace SimpleDateFormat with DateTimeFormatter of Java 8

fuwei123456:
SimpleDateFormat is not thread safe


zilm13:
Looks good, thanks!

PR #1073, opened 2018-05-11
Added verifying of encoded RLP length to be not greater than available data size

zilm13:
Verifying of decoded RLP object length to be not greater than data size

While primary decode method, `RLP.decode2(byte[])` is protected to verify whether decoded length lies in the boundaries of remaining data length, we have several other public methods which doesn't include such inspection. One of them, `RLP.decode(byte[], int)` is used, for example, in `Value` constructor and could be attacked using incorrect RLP data. Length is int, so maximum possible size is 2+B or more than 2Gb occupied by each RLPItem from attacker's RLP. So this kind of attack could occupy all available memory.

What's done:
- test added with data from such kind of attack
- added verification to all RLP class methods which could occupy more than 1 byte of length of length


ethereumj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -694,6 +694,19 @@ private static void fullTraverse(byte[] msgData, int level, int startPos,
...
+    /**
+     * Compares supplied length information with maximum possible
+     * @param suppliedLength    Length info from header
+     * @param availableLength   Length of remaining object
+     * @throws RuntimeException if supplied length is bigger than available
+     */
+    private static void verifyLength(int suppliedLength, int availableLength) {
+        if (suppliedLength > availableLength) {
+            throw new RuntimeException(String.format("Length parsed from RLP (%s bytes) is greater " +
+                            "than remaining size of data (%s bytes)", suppliedLength, availableLength));
```
mkalinin: I think it'd be better to read something like `than [probable | possible] size of data`

zilm13:
> I'd also add that check to this method: https://github.com/ethereum/ethereumj/pull/1073/files#diff-424bd3538481cd5c9df18504819120b6R776

this one is encoding 
plus length is limited to one byte (less than 256)

PR #1074, opened 2018-05-11
Several fixes for FastSync

zilm13:
- Headers DB removed after fast sync is over if "skipHistory" option was not enabled. No restart is required
- Fixed bug in `SyncQueueReverseImpl` which could force Fast sync to stall for a long time when it needs only to download 256 blocks
- More aggressive sync pool strategy is used during syncing: other peers are dropped if they are not used and we need more peers for sync while other peers are not suitable for our needs


ethereumj-core/src/main/java/org/ethereum/net/server/EthereumChannelInitializer.java
```diff
@@ -57,11 +62,35 @@ public void initChannel(NioSocketChannel ch) throws Exception {
...
+                    logger.debug("Drop connection - bad remote address, channel: {}", ch.toString());
+                    needToDrop = true;
+                }
+                // Avoid too frequent connection attempts
+                if (!needToDrop && channelManager.isRecentlyDisconnected(ch.remoteAddress().getAddress())) {
+                    logger.debug("Drop connection - the same IP was disconnected recently, channel: {}", ch.toString());
+                    needToDrop = true;
+                }
+                // Drop bad peers before creating channel
+                if (!needToDrop && nodeManager.getNodeStatistics(new Node(new byte[0], ch.remoteAddress().getHostString(),
```
mkalinin: Requesting for `nodeManager.isReputationPenalized(InetSocketAddress addr)` which does all the job

ethereumj-core/src/main/java/org/ethereum/sync/FastSyncManager.java
```diff
@@ -409,11 +412,14 @@ public void onSuccess(List<Pair<byte[], byte[]>> result) {
...
                             synchronized (FastSyncManager.this) {
                                 logger.trace("Received " + result.size() + " nodes (of " + hashes.size() + ") from peer: " + idle);
+                                idle.getNodeStatistics().eth63NodesRequested.add(hashes.size());
+                                idle.getNodeStatistics().eth63NodesRetrieveTime.add(System.currentTimeMillis() - reqTime);
                                 for (Pair<byte[], byte[]> pair : result) {
                                     TrieNodeRequest request = pendingNodes.get(pair.getKey());
                                     if (request == null) {
                                         long t = System.currentTimeMillis();
                                         logger.debug("Received node which was not requested: " + toHexString(pair.getKey()) + " from " + idle);
+                                        idle.disconnect(ReasonCode.USELESS_PEER);
```
mkalinin: This case could not be related to the meaning of log message standing above. It may be a result of  `pendingNodes` purging due to `REQUEST_TIMEOUT`. Isn't it too rude to disconnect peer with `USELESS_PEER` reason in that case? Btw, log message could also be fixed to avoid misunderstanding in the future.

PR #1080, opened 2018-05-21
Issue #1078 Gradle Task to create executable fatJar

### LINKED ISSUE #1078: Generated jar files are not executable
on running jar file,
```
java -jar ethereumj-core/build/libs/ethereumj-core-all-1.8.0-SNAPSHOT.jar
```
I always get 
```
no main manifest attribute
```

kishansagathiya:
This commit adds a gradle task which will create an executable jar
with all the dependencies

Fixes #1078

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ git clone https://github.com/ethereum/ethereumj
 cd ethereumj
 cp ethereumj-core/src/main/resources/ethereumj.conf ethereumj-core/src/main/resources/user.conf
 vim ethereumj-core/src/main/resources/user.conf # adjust user.conf to your needs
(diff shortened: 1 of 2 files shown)


kishansagathiya:
@mkalinin 
I was able to create an executable jar without using a dependency like Shadow.
Eventhough, I have raised this PR, I have a feeling that you either don't want a task to create executable jar( because you removed shadowJar). In which case, feel free to close this.
Any reason for removing shadowJar? Does this PR help?

zilm13:
@kishansagathiya No need to do everything in one commit next time. It's better to see changes in separate commit according to review and not a big deal if we have 2+ commits for one pr.

PR #1082, opened 2018-05-24
Refuse connection from ips that are already in processing queue

zilm13:
@mkalinin please, take a look

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
+++ b/ethereumj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
@@ -205,7 +205,7 @@ public class ChannelManager {
      * @param peerAddr      Peer address
      * @return true if we already have connection from this address, otherwise false
      */
-    public boolean ipAlreadyWaiting(InetAddress peerAddr) {
+    public boolean isAddressInQueue(InetAddress peerAddr) {
         for (Channel peer: newPeers) {
             if (peer.getInetSocketAddress().getAddress().getHostAddress().equals(peerAddr.getHostAddress())) {
                 return true;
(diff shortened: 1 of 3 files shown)


ethereumj-core/src/main/java/org/ethereum/net/server/EthereumChannelInitializer.java
```diff
@@ -115,6 +92,43 @@ public void initChannel(NioSocketChannel ch) throws Exception {
             logger.error("Unexpected error: ", e);
         }
     }
+    
+    private boolean shouldRefuseIncoming(NioSocketChannel ch) {
+        if(!isInbound()) return false;
```
mkalinin: it's better to check it outside

ethereumj-core/src/main/java/org/ethereum/net/server/EthereumChannelInitializer.java
```diff
@@ -115,6 +92,43 @@ public void initChannel(NioSocketChannel ch) throws Exception {
             logger.error("Unexpected error: ", e);
         }
     }
+    
+    private boolean shouldRefuseIncoming(NioSocketChannel ch) {
```
mkalinin: Maybe something like `isEligibleForConnection`?

ethereumj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
```diff
@@ -200,6 +200,21 @@ public void disconnect(Channel peer, ReasonCode reason) {
         recentlyDisconnected.put(peer.getInetSocketAddress().getAddress(), new Date());
     }
 
+    /**
+     * Whether peer with the same ip is in newPeers, waiting for processing
+     * @param peerAddr      Peer address
+     * @return true if we already have connection from this address, otherwise false
+     */
+    public boolean ipAlreadyWaiting(InetAddress peerAddr) {
```
mkalinin: `isAddressInQueue`?

ethereumj-core/src/main/java/org/ethereum/net/server/EthereumChannelInitializer.java
zilm13: i like it!

zilm13: almost )

PR #1084, opened 2018-05-27
Added events in Miner listener on DAG generation start and end

zilm13:
DAG generation, especially full DAG generation could take a lot of time, and the only way currently to detect its start and end is to parse logs, which is ehhhhh. 
So I've added event for it anyone could use with `MinerListener`

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
+++ b/ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
@@ -259,15 +259,14 @@ public class BlockMiner {
             miningBlock = newMiningBlock;
 
             if (externalMiner != null) {
+                externalMiner.setListeners(listeners);
                 currentMiningTasks.add(externalMiner.mine(cloneBlock(miningBlock)));
             }
             if (isLocalMining) {
                 MinerIfc localMiner = config.getBlockchainConfig()
                         .getConfigForBlock(miningBlock.getNumber())
                         .getMineAlgorithm(config);
-                if (localMiner instanceof EthashMiner) {
-                    ((EthashMiner) localMiner).setListeners(listeners);
-                }
+                localMiner.setListeners(listeners);
                 currentMiningTasks.add(localMiner.mine(cloneBlock(miningBlock)));
             }
 
--- a/ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
+++ b/ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
@@ -38,6 +38,7 @@ import org.springframework.stereotype.Component;
 import java.math.BigInteger;
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.stream.Collectors;
 
 import static java.lang.Math.max;
 
@@ -265,7 +266,13 @@ public class BlockMiner {
                 MinerIfc localMiner = config.getBlockchainConfig()
                         .getConfigForBlock(miningBlock.getNumber())
                         .getMineAlgorithm(config);
-                localMiner.setListeners(listeners);
+                if (localMiner instanceof EthashMiner) {
+                    Set<EthashMinerListener> ethashMinerListeners = listeners.stream()
+                            .filter(listener -> listener instanceof EthashMinerListener)
+                            .map(listener -> (EthashMinerListener) listener)
+                            .collect(Collectors.toSet());
+                    ((EthashMiner) localMiner).setListeners(ethashMinerListeners);
+                }
                 currentMiningTasks.add(localMiner.mine(cloneBlock(miningBlock)));
             }
 
--- a/ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
+++ b/ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
@@ -38,7 +38,6 @@ import org.springframework.stereotype.Component;
 import java.math.BigInteger;
 import java.util.*;
 import java.util.concurrent.*;
-import java.util.stream.Collectors;
 
 import static java.lang.Math.max;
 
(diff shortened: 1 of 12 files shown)


mkalinin:
`listeners` instance shared across `BlockMiner`, `EthHashMiner` and `Ethash` looks a bit messy. What if we localize `EthashMinerListener extends MinerListener`? 
Then it can be possible to manage a list of Ethash listeners in `EthHashMiner` passing them throughout `BlockMiner`. The other point of the inheritance is forward compatibility, users just won't have to update their listeners with new event.

ethereumj-core/src/main/java/org/ethereum/mine/MinerListener.java
```diff
@@ -23,9 +23,30 @@
  * Created by Anton Nashatyrev on 10.12.2015.
  */
 public interface MinerListener {
+
+    enum MinerStatus {
```
mkalinin: What is valuable for the user of that status? I suppose the most valuable things to be notified about are starting and ending of dataset creation. Concerning an option to mine with usage of light dataset only, this lifecycle not always looks perfect. I think that better to have following cycle: `LIGHT_STARTED` -> `FULL_STARTED (optionally)` -> `DATASET_GENERATED`. Last message could have two options `DATASET_GENERATED_LIGHT` and `DATASET_GENERATED_FULL`

ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
```diff
@@ -265,7 +266,13 @@ protected void restartMining() {
                 MinerIfc localMiner = config.getBlockchainConfig()
                         .getConfigForBlock(miningBlock.getNumber())
                         .getMineAlgorithm(config);
-                localMiner.setListeners(listeners);
+                if (localMiner instanceof EthashMiner) {
+                    Set<EthashMinerListener> ethashMinerListeners = listeners.stream()
```
mkalinin: Maybe it would be better to do filtering in `EthashMiner.setListeners()` to keep the main flow clean?

ethereumj-core/src/main/java/org/ethereum/mine/Ethash.java
```diff
@@ -78,12 +82,21 @@ public static Ethash getForBlock(SystemProperties config, long blockNumber) {
...
-    public static Ethash getForBlock(SystemProperties config, long blockNumber, List<MinerListener> listeners) {
+    public static Ethash getForBlock(SystemProperties config, long blockNumber, Collection<EthashMinerListener> listeners) {
         Ethash ethash = getForBlock(config, blockNumber);
-        ethash.listeners = listeners;
+        ethash.listeners.clear();
+        ethash.listeners.addAll(listeners);
         return ethash;
     }
 
+    public boolean addListener(EthashMinerListener listener) {
```
mkalinin: Let's remove them both if they are not used

ethereumj-core/src/main/java/org/ethereum/mine/EthashMiner.java
```diff
@@ -56,8 +57,28 @@ public boolean validate(BlockHeader blockHeader) {
...
+    public void setListeners(Collection<EthashMinerListener> listeners) {
+        this.listeners.clear();
+        this.listeners.addAll(listeners);
+    }
+
+    /**
+     * Listeners changes affects only future {@link #mine(Block)} and
+     * {@link #validate(BlockHeader)} calls
+     */
+    public boolean addListener(EthashMinerListener listener) {
```
mkalinin: Same as above, let's remove these two methonds

ethereumj-core/src/main/java/org/ethereum/mine/EthashMinerListener.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with the ethereumJ library. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.ethereum.mine;
+
+/**
+ * {@link MinerListener} designed for use with {@link EthashMiner}
+ */
+public interface EthashMinerListener extends MinerListener {
```
mkalinin: I'd cut the name to `EthashListener`

ethereumj-core/src/main/java/org/ethereum/mine/BlockMiner.java
```diff
@@ -265,6 +265,9 @@ protected void restartMining() {
                 MinerIfc localMiner = config.getBlockchainConfig()
                         .getConfigForBlock(miningBlock.getNumber())
                         .getMineAlgorithm(config);
+                if (localMiner instanceof EthashMiner) {
```
mkalinin: sorry, but it looks really close to what we had in the first iteration 🤣  Maybe bring back that `setListeners` method to the interface and remove type checking?

pschweitz:
Hi,

Thanks for your job!

I would go a bit further, I think it could be relevant to trigger the event after the load from the file system as well, for both DAG (line 131 and 174 of Ethash.java), and distinguish which has been generated thanks to another new Status: 

DATASET_GENERATED -> renamed to FULL_DATASET_GENERATED 
LIGHT_DATASET_GENERATED

zilm13:
@pschweitz the idea behind this improvement was to track long event delaying block mining - dataset generation. Loading it from disk takes not much time and old events like startMining looks enough for it. We don't want to push event `LIGHT_DATASET_GENERATED ` in case when there is ongoing full dataset generation, because this event do not give you info about ongoing delay. You couldn't be sure whether full dataset generation is going or not. Current approach sends `GENERATED` info only once so you could be sure, that block mining will not be delayed anymore after receiving it.

zilm13:
@pschweitz we've discussed it again with @mkalinin and we think it's a good time to change it until its going to release. Also you have some good points!
You could join discussion at #1091

PR #1087, opened 2018-05-29
Fix/GitHub tests

zilm13:
Cache made with properties file and storing downloaded files. According to log it's used on travis instead of download. We could try this solution. What do you think?


zilm13:
@mkalinin how about such approach?
It will be difficult to parse `test-ethereumj.conf` in gradle config because we are using typesafe Config format, not native Java properties. I guess it's doable, but should we?

mkalinin:
The incentive behind storing it in the config is to avoid frequent updates of `.travis.yml` and keep user away from passing commit hash when it calls `prepareGithubTests`.

zilm13:
@mkalinin review me, plz

ethereumj-core/build.gradle
```diff
@@ -356,3 +368,73 @@ def gitCurBranch() {
...
+            testsRepo.checkout(branch: commit)
+        } catch (Exception ex) {
+            println "Error occurs while trying to checkout local Ethereum tests repo to specific commit"
+            println "Either remove " + testsPathKey + " property to use remote repo in tests or fix the issue"
+            println "Error: "
+            println ex
+            throw ex
+        }
+    }
+}
```
mkalinin: new line pls

ethereumj-core/src/test/resources/github-tests.prop
```diff
@@ -0,0 +1,5 @@
+# Used by Gradle task prepareGithubTests
+# to determinate commit of Ethereum tests repo https://github.com/ethereum/tests
+# which should be checked out in local repo directory
+# EthereumJ is tested with files from this commit
+GitHubTests.commit=7f638829311dfc1d341c1db85d8a891f57fa4da7
```
mkalinin: new line pls

PR #1098, opened 2018-06-14
[WIP] Limit addPendingTransactions lock time

### LINKED ISSUE #1076: Limit addPendingTransactions lock time
`addPendingTransactions` could lock `PendingState` object for pretty long time if e.g. 30k transactions have come from the network, therefore `processBest` methods stall for that period of time and finally `PendingState` lags.
Possible solution is to introduce buffered queue for transaction processing to let `processBest` acquire `PendingState` object lock. It should be a LIFO queue to keep track of most fresh data got from the net and it also must have a cap on its size to prevent high memory consumption.

kishansagathiya:
Clearly, this does not fix the issue since tests are not passing.
@mkalinin @zilm13 You can take a look at it and let me know if I am doing it right.

Fixes #1076

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/AccountState.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/AccountState.java
@@ -19,7 +19,6 @@ package org.ethereum.core;
 
 import org.ethereum.config.BlockchainConfig;
 import org.ethereum.config.SystemProperties;
-import org.ethereum.crypto.HashUtil;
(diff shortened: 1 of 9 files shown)


mkalinin:
@kishansagathiya changes you have made does not address the issue. `addPendingTransactions` method still processes all passed transactions in one call that locks `PendingState` for execution time.

kishansagathiya:
It's been quite some time for this, I will have start from the beginning if I were to finish. Closing this so that someone else can pick up the issue.

PR #1104, opened 2018-06-15
#972: Strategy for generating nodeId and nodePrivateKey

### LINKED ISSUE #972: Strategy interface for getGeneratedNodePrivateKey()
There is a TODO on the test `testGeneratedNodePrivateKeyThroughECKey` on the class `SystemPropertiesTest` to create a strategy interface for getGeneratedNodePrivateKey().

```
/**
 * TODO: Consider using a strategy interface for #getGeneratedNodePrivateKey().
 * Anything 'File' and 'random' generation are difficult to test and assert
 */
@Test
public void testGeneratedNodePrivateKeyThroughECKey() throws Exception {
    ...
}
```

I'd love to get someone to review my PR and come with suggestions for improvements.

lucassaldanha:
Fixes #972

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/config/GenerateNodeIdRandomly.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/GenerateNodeIdRandomly.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) [2017] [ <ether.camp> ]
+ * This file is part of the ethereumJ library.
+ *
+ * The ethereumJ library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The ethereumJ library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


zilm13:
@lucassaldanha I'd prefer direct way to set fallback strategy in `GenerateNodeIdFromPropsFile` and change its name to `GetNodeIdFromPropsFile`. You will need to pass databaseDir 2 times in `SystemProperties` on init but it will not look like a hidden magic. Everything else looks great!
@mkalinin what do you think?

zilm13:
So it will look this way in SystemProperties:
```
this.generateNodeIdStrategy = new GetNodeIdFromPropsFile(databaseDir()).withFallback(new GenerateNodeIdRandomly(databaseDir()));
```

lucassaldanha:
> I'd prefer direct way to set fallback strategy in GenerateNodeIdFromPropsFile

@zilm13 Sure! Do you want the fallback to be optional? e.g. If one doesn't specify the fallback, what should happen? Thrown an error?

zilm13:
@lucassaldanha yeah I think exception is better than null, so we get something wrong immediately if we have bad initialization. And you should be able to pass any `GenerateNodeIdStrategy ` in fallback.

zilm13:
@lucassaldanha Looks good, thanks

PR #1105, opened 2018-06-18
Prevent early switch to SyncState.COMPLETE

mkalinin:


zilm13:
@mkalinin Looks good to me

PR #1110, opened 2018-06-25
Set NameOnlyOptions as user-based parameters and add GAS option

tbocek:
With this pull request, a user can specify e.g., --gas to return the estimated gas costs. To remain
backwards compatible, optimize flag is still in place. If optimize flag is set, the option optimize will not be added twice

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/solidity/compiler/SolidityCompiler.java
+++ b/ethereumj-core/src/main/java/org/ethereum/solidity/compiler/SolidityCompiler.java
@@ -63,8 +63,9 @@ public class SolidityCompiler {
         public static final OutputOption METADATA = OutputOption.METADATA;
         public static final OutputOption ASTJSON = OutputOption.ASTJSON;
 
-        private static final NameOnlyOption OPTIMIZE = NameOnlyOption.OPTIMIZE;
-        private static final NameOnlyOption VERSION = NameOnlyOption.VERSION;
+        public static final NameOnlyOption OPTIMIZE = NameOnlyOption.OPTIMIZE;
+        public static final NameOnlyOption VERSION = NameOnlyOption.VERSION;
+        public static final NameOnlyOption GAS = NameOnlyOption.GAS;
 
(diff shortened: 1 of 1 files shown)


zilm13:
@tbocek you cannot use VERSION along with anything else. I think it's better to put GAS to `OutputOption` and revert all other changes. Though gas estimation will not be in json format like everything else. But it's part of output and it works ok along with other options.

tbocek:
@zilm13 Having GAS as an OutputOption does not work as the following:
```solc --combined-json bin,abi,gas test.sol```
returns: ```Invalid option to --combined-json: gas```
So, it has to be ```solc --combined-json bin,abi --gas test.sol```

I can revert VERSION, but I think GAS and OPTIMIZE should be exposed to the user. What do you think?

**Comment by zilm13** [comment]
@tbocek @gsmachado
How about something like that?

```

    public static class CustomOption implements Option {
        private String name;
        private String value;

        public CustomOption(String name) {
            if (name.startsWith("--")) {
                this.name = name.substring(2);
            } else {
                this.name = name;
            }
        }

        public CustomOption(String name, String value) {
            this(name);
            this.value = value;
        }

        @Override
        public String getValue() {
            return value;
        }

        @Override
        public String getName() {
            return name;
        }
    }

[...]

        for (Option option : getElementsOf(CustomOption.class, options)) {
            commandParts.add("--" + option.getName());
            if (option.getValue() != null) {
                commandParts.add(option.getValue());
            }
        }
```

So we could use something like that:

```
        SolidityCompiler.Result res = SolidityCompiler.compile(
                contractSrc.getBytes(), true, ABI, BIN, INTERFACE, METADATA, new SolidityCompiler.CustomOption("gas"));
```

So user have options not only for gas and we don't need to use `NameOnlyOption` with hacks both for `version` and `optimize`.

tbocek:
makes sense. I have updated the pull request accordingly

zilm13:
@tbocek looks good, thanks!

PR #1120, opened 2018-07-05
Fix/rc active peers

mkalinin:


ethereumj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
```diff
@@ -153,7 +153,7 @@ public void connect(Node node) {
         return ids;
     }
 
-    private void processNewPeers() {
+    private synchronized void processNewPeers() {
```
Nashatyrev: Such things like `channel.disconnect()` under lock make me nervous :(
I.e. we are calling method which can potentially traverse many components. This is very deadlock prone approach imho

ethereumj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
```diff
@@ -344,7 +344,7 @@ public void add(Channel peer) {
         newPeers.add(peer);
     }
 
-    public void notifyDisconnect(Channel channel) {
+    public synchronized void notifyDisconnect(Channel channel) {
         logger.debug("Peer {}: notifies about disconnect", channel);
         channel.onDisconnect();
```
Nashatyrev: The same here. There is no need to call those notify methods under the lock. They can potentially have deadlock side effects

Nashatyrev:
IMHO synchronizing such central manager class should be done very carefully. 
Don't like outer calls under locks in the original fix. They are potentially deadlock prone. Even though the current implementation may not deadlock minor harmless change on the other side may cause the problem. 
The only thing we need to assure here is that on `notifyDisconnect` call the peer would be 100% deleted from either `newPeers` or `activePeers`. To achieve this we would better synchronize only those fields access

ethereumj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
```diff
@@ -154,44 +154,48 @@ public void connect(Node node) {
...
                     }
-                } else {
-                    disconnect(peer, DUPLICATE_PEER);
+
+                    processed.add(peer);
                 }
+            }
 
-                processed.add(peer);
+            if (addCnt > 0) {
```
zilm13: `addCnt` is not used anymore in this code

PR #1122, opened 2018-07-07
Created VMHookFactory

### LINKED ISSUE #1070: Refactor VMHook
VM hooks are currently not thread safe due to sharing one `VMHook` instance across all `VM` instances. For example, it turns into concurrent modifications because of `PendingState` processing. A straightforward way of solving this issue is to instantiate own `VMHook` instance for each `VM`. One of the way to achieve this is to introduce `VMHookFactory`

RichardH92:
Closes #1070

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
@@ -157,7 +157,6 @@ public class SystemProperties {
     private Genesis genesis;
     private Boolean vmTrace;
     private Boolean recordInternalTransactionsData;
-    private VMHook vmHook;
 
     private final ClassLoader classLoader;
 
@@ -639,11 +638,6 @@ public class SystemProperties {
     @ValidateMe
     public boolean vmTrace() {
         return vmTrace == null ? (vmTrace = config.getBoolean("vm.structured.trace")) : vmTrace;
-    }
-
-    @ValidateMe
-    public VMHook vmHook() {
-        return null;
     }
 
     @ValidateMe
--- a/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
@@ -37,7 +37,6 @@ import org.ethereum.util.ByteUtil;
 import org.ethereum.util.Utils;
 import org.ethereum.validator.BlockCustomHashRule;
 import org.ethereum.validator.BlockHeaderValidator;
-import org.ethereum.vm.VMHook;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.spongycastle.util.encoders.Hex;
--- a/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
@@ -37,6 +37,7 @@ import org.ethereum.util.ByteUtil;
 import org.ethereum.util.Utils;
 import org.ethereum.validator.BlockCustomHashRule;
 import org.ethereum.validator.BlockHeaderValidator;
+import org.ethereum.vm.VMHook;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.spongycastle.util.encoders.Hex;
@@ -156,6 +157,7 @@ public class SystemProperties {
     private Genesis genesis;
     private Boolean vmTrace;
     private Boolean recordInternalTransactionsData;
+    private VMHook vmHook;
 
     private final ClassLoader classLoader;
 
@@ -637,6 +639,11 @@ public class SystemProperties {
     @ValidateMe
     public boolean vmTrace() {
         return vmTrace == null ? (vmTrace = config.getBoolean("vm.structured.trace")) : vmTrace;
+    }
+
+    @ValidateMe
+    public VMHook vmHook() {
+        return null;
     }
 
     @ValidateMe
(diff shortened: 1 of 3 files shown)


ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1328,8 +1333,8 @@ public void play(Program program) {
         }
     }
 
-    public static void setVmHook(VMHook vmHook) {
-        VM.vmHook = vmHook;
+    public static void setVmHookFactory(VMHookFactory vmHookFactory) {
```
mkalinin: Let's keep old `VMHook` behavior for backward compatibility and mark it as deprecated with pointing to new alternative with detailed description of why client should use new feature

ethereumj-core/src/main/java/org/ethereum/vm/VMHookFactory.java
```diff
@@ -0,0 +1,8 @@
+package org.ethereum.vm;
+
+/**
+ * Created by Richard Hart on 07.07.2018.
+ */
+public interface VMHookFactory {
```
mkalinin: Please, provide a description of what this class is needed for

ethereumj-core/src/main/java/org/ethereum/vm/VMHookFactory.java
```diff
@@ -0,0 +1,8 @@
+package org.ethereum.vm;
```
mkalinin: Lack of license header

ethereumj-core/src/main/java/org/ethereum/vm/VMHookFactory.java
```diff
@@ -0,0 +1,8 @@
...
+
+/**
+ * Created by Richard Hart on 07.07.2018.
+ */
+public interface VMHookFactory {
+    VMHook build();
```
mkalinin: Let's use `create` naming for this method.

RichardH92:
@mkalinin Thanks for the feedback! I think this is ready for re-review, if you don't mind.

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -367,6 +374,8 @@ else if (oldValue != null && newValue.isZero()) {
 
             if (vmHook != null) {
                 vmHook.step(program, op);
+            } else if (globalVmHook != null) {
```
mkalinin: `globalVmHook` must be triggered even if `vmHookFactory` is defined

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -20,6 +20,7 @@
 import org.ethereum.config.BlockchainConfig;
 import org.ethereum.config.SystemProperties;
 import org.ethereum.db.ContractDetails;
+import org.ethereum.sync.FastSyncManager;
```
mkalinin: Is that dependency needed here?

ethereumj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -37,6 +37,7 @@
 import org.ethereum.util.Utils;
 import org.ethereum.validator.BlockCustomHashRule;
 import org.ethereum.validator.BlockHeaderValidator;
+import org.ethereum.vm.VMHook;
```
mkalinin: Looks like redundant dependency

RichardH92:
Hey @mkalinin I believe I have addressed your 2nd round of comments. Please lmk if I need to change anything else

zilm13:
Looks good to me, but I'd also like to see vmHooks calls to be bounded with try/catch so vmHook couldn't fail VM execution or is it excess protection, @mkalinin

RichardH92:
@zilm13 That's a good idea. Just to be clear, you want me to catch all exceptions, right? Since this will be implemented by a third party? Should I just log the stacktrace?

zilm13:
@RichardH92 yeah, just try-catch-log.error("... {}", exception), separate contours for `vmHook` and `globalVmHook` as we are going to remove last with time and to exclude interfere.

RichardH92:
Hey @zilm13 how does this look to you? Any suggestions on the error messages?

zilm13:
@RichardH92 How about adding something like this?
1) Don't like duplicates
2) You are not logging exceptions
I think it could be fixed with something like this:
```
    private void callVmHookAction(Program program, BiConsumer<VMHook, Program> action) {
        if (vmHook != null) {
            try {
                action.accept(vmHook, program);
            } catch (Exception e) {
                logger.error("Error calling VMHook action: {}", e);
            }
        }

        if (globalVmHook != null) {
            try {
                action.accept(globalVmHook, program);
            } catch (Exception e) {
                logger.error("Error calling global VMHook action: {}", e);
            }
        }
    }
```

and then call it:
```
callVmHookAction(program, (hook, prg) -> hook.step(prg, op));
callVmHookAction(program, VMHook::startPlay);
callVmHookAction(program, VMHook::stopPlay);
```

also, please, add exception logging to `.create` with smth like this:
```
logger.error("Error creating VMHook: {}", e);
```

RichardH92:
Hey @zilm13 , thanks for the feedback. I think this is ready for re-review when you have time

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -375,7 +376,7 @@ else if (oldValue != null && newValue.isZero()) {
             if (program.getNumber().intValue() == dumpBlock)
                 this.dumpLine(op, gasBefore, gasCost + callGas, memWords, program);
 
-            vmHookStepWrapper(program, op);
+            callVmHookAction(program, (hook, prg) -> vmHook.step(prg, op));
```
zilm13: `(hook, prg) -> vmHook.step(prg, op)` should be changed to `(hook, prg) -> hook.step(prg, op)` or you will call it twice in wrapper

RichardH92:
@zilm13 Nice catch! Should be fixed now

zilm13:
@RichardH92 Cool, thx!

PR #1125, opened 2018-07-11
Fix: ethash rule

mkalinin:
Not for being merged yet. Just for first glance review.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
+++ b/ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
@@ -111,6 +111,8 @@ public interface EthereumListener {
 
     void onBlock(BlockSummary blockSummary);
 
+    void onBlock(BlockSummary blockSummary, boolean best);
+
     void onPeerDisconnect(String host, long port);
 
     /**
--- a/ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
+++ b/ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
@@ -111,7 +111,9 @@ public interface EthereumListener {
 
     void onBlock(BlockSummary blockSummary);
 
-    void onBlock(BlockSummary blockSummary, boolean best);
+    default void onBlock(BlockSummary blockSummary, boolean best) {
+        onBlock(blockSummary);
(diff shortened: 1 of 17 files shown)


ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
```diff
@@ -111,6 +111,8 @@ public boolean isPending() {
 
     void onBlock(BlockSummary blockSummary);
 
+    void onBlock(BlockSummary blockSummary, boolean best);
+
```
Nashatyrev: You should better add
```java
default void onBlock(BlockSummary blockSummary, boolean best) {
    onBlock(blockSummary);
}
```
Else you will break existing code depending on EthJ

ethereumj-core/src/main/java/org/ethereum/listener/EthereumListenerAdapter.java
```diff
@@ -44,6 +44,11 @@ public void onBlock(BlockSummary blockSummary) {
         onBlock(blockSummary.getBlock(), blockSummary.getReceipts());
     }
 
+    @Override
+    public void onBlock(BlockSummary blockSummary, boolean best) {
+        onBlock(blockSummary);
+    }
+
```
Nashatyrev: Remove this method if add interface `default`

ethereumj-core/src/main/java/org/ethereum/mine/EthashValidationHelper.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    /**
+     * Calculates ethash results for particular block and nonce.
+     *
+     * @param cachedOnly flag that defined behavior of method when dataset has not been cached:
+     *                   if set to true  - returns null immediately
+     *                   if set to false - generates dataset for block epoch and then runs calculations on it
+     */
+    public Pair<byte[], byte[]> ethashWorkFor(BlockHeader header, byte[] nonce, boolean cachedOnly) throws Exception {
```
Nashatyrev: I would revisit synchronization on two public methods here, since they could be called concurrently

mkalinin:
I think it's ready for review and merge now

PR #1132, opened 2018-07-17
Fix indention and space

skyline75489:


mkalinin:
We don't make big code formatting changes. It's better to introduce them within functional updates.

PR #1134, opened 2018-07-22
modify gradle config

chenjw13097:


zilm13:
Why should we change this?

PR #1135, opened 2018-07-22
Nested lists rlp bomb fix

zilm13:
Please, check


ethereumj-core/src/test/java/org/ethereum/util/RLPTest.java
```diff
@@ -1215,4 +1215,536 @@ public void rlpEncodedLength() {
...
+            RLP.decode2(data);
+        } catch (RuntimeException ex) {
+            Throwable cause = ex;
+            while (cause != null) {
+                if (cause.getMessage().contains("big") && cause.getMessage().contains("nesting")) tooBigNestedFired = true;
+                cause = cause.getCause(); // To the next level
+            }
+        }
+
+        assert tooBigNestedFired;
```
mkalinin: Maybe just use `fail()` instead?

ethereumj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -547,20 +550,25 @@ public static byte getCommandCode(byte[] data) {
...
     public static RLPElement decode2OneItem(byte[] msgData, int startPos) {
         RLPList rlpList = new RLPList();
-        fullTraverse(msgData, 0, startPos, startPos + 1, 1, rlpList);
+        fullTraverse(msgData, 0, startPos, startPos + 1, 1, rlpList, new AtomicInteger(1));
         return rlpList.get(0);
     }
     /**
      * Get exactly one message payload
      */
     private static void fullTraverse(byte[] msgData, int level, int startPos,
```
mkalinin: What if use `level` argument instead of adding a new one?

zilm13:
Here are the set of changes:
- Instead of complex nested counter just throwing Exception on depth higher than 16
- Some strange duplicated copy of `fullTraverse` removed
- Added methods for `RLP.decode2` with depth limit
- All usages of RLP.decode2, where full decode is not really used was modified to use decode with limited depth

ethereumj-core/src/main/java/org/ethereum/manager/BlockLoader.java
```diff
@@ -112,7 +112,7 @@ public void loadBlocks() {
                 Path path = Paths.get(fileSrc);
                 // NOT OPTIMAL, but fine for tests
                 byte[] data = Files.readAllBytes(path);
-                RLPList list = RLP.decode2(data);
+                RLPList list = RLP.decode2ListNarrow(data);
```
mkalinin: It turns that this update changes the logic: `RLP.decode2(data, Integer.MAX)` is replaced with `RLP.decode2(data, 2).get(0)`. Additional `get(0)` is added in the latter statement.

ethereumj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -547,20 +461,33 @@ public static byte getCommandCode(byte[] data) {
...
         return rlpList;
     }
 
+    /**
+     * Decodes without going deep after 1st level list
+     * (actually, 2nd as RLP nesting one extra level)
+     * @param msgData rlp data
+     * @return list of RLPItems
+     */
+    public static RLPList decode2ListNarrow(byte[] msgData) {
```
mkalinin: Let's use `unwrapList` name for this method. Short and addresses its meaning pretty well

mkalinin: I'd also update description with an example of how does it work.

ethereumj-core/src/main/java/org/ethereum/manager/BlockLoader.java
zilm13: Yeah, I've checked Hive benchmarks this method was added for, they use this kind of list wrapping. Fixed

mkalinin:
Looks good now 👍 
I'd merge it if you say that changes are well tested

Issue #1127, opened 2018-07-13
Registration of beacon chain validators


mkalinin:
Adopt [registration contract](https://github.com/ethereum/beacon_chain/blob/master/contracts/validator_registration.v.py) and implement logic to manage validators set

PR #1137, opened 2018-07-26
Bootstrap validator reg contract

mkalinin:
Just for review, not for merge yet! 
Resolves https://github.com/ethereum/ethereumj/issues/1127


ethereumj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -290,6 +291,30 @@ public synchronized long getMaxNumber(){
...
+        long qty = to.getNumber() - from.getNumber() + 1;
+
+        List<Block> ret = new ArrayList<>();
+        Block block = to;
+        for (long i = 0; i < qty && block != null && !from.isEqual(block); ++i) {
+            ret.add(block);
+            block = blocks.get(block.getParentHash());
+        }
+
+        if (ret.size() > 0 && !ret.get(ret.size() - 1).isEqual(from))
```
zilm13: why this could happens?

ethereumj-core/src/main/java/org/ethereum/db/BlockStore.java
```diff
@@ -49,6 +49,17 @@
...
+    /**
+     * Returns a list of blocks starting from toHash (inclusively) down to fromHash (inclusively).
+     * Both hashes should be of blocks belonging the same chain, otherwise method returns an empty list.
+     * If some blocks between from and to are missed then empty list is returned.
+     *
+     * @param fromHash lower bound of a list
+     * @param toHash upper bound of a list
+     * @return returns a list with inverted block order
+     */
+    List<Block> listBlocks(byte[] fromHash, byte[] toHash);
```
zilm13: I'd focus backwards nature of method in it's name or input param names

ethereumj-core/src/main/java/org/ethereum/sharding/ShardingEntry.java
```diff
@@ -0,0 +1,64 @@
...
+import java.net.URISyntaxException;
+
+/**
+ * @author Mikhail Kalinin
+ * @since 26.07.2018
+ */
+public class ShardingEntry {
+
+    public static void main(String args[]) throws IOException, URISyntaxException {
+        System.setProperty("ethereumj.conf.res", "sharding.conf");
```
zilm13: btw I think it will be good feature to let user pass several configs
and this way we will not loose any user config already passed in such parameter

ethereumj-core/src/main/java/org/ethereum/sharding/config/BeaconConfig.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @author Mikhail Kalinin
+ * @since 21.07.2018
+ */
+@Configuration
+@Conditional(BeaconConfig.Enabled.class)
+public class BeaconConfig {
+
+    @Autowired
```
zilm13: It's very dangerous to use `@Autowired` in configuration, but if it works for you and it's prototype we could keep it

ethereumj-core/src/main/java/org/ethereum/sharding/domain/Validator.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private Validator() {
+    }
+
+    public static Validator fromLogData(byte[] data) {
+        if (data.length < LOG_DATA_SIZE)
+            return null;
+
+        Validator v = new Validator();
+        v.pubKey = copyOfRange(data, 0, 32);
```
zilm13: Couldn't we use something like this https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/core/CallTransaction.java#L292 for initial parsing here? Current thing is like a magic

ethereumj-core/src/main/java/org/ethereum/sharding/contract/DepositContract.java
```diff
@@ -0,0 +1,153 @@
...
+    RecommendedGasPriceTracker gasPriceTracker;
+
+    CompletableFuture<TransactionReceipt> depositFuture;
+    byte[] depositTxHash;
+
+    public DepositContract(Ethereum ethereum, byte[] address, String abi) {
+        this.ethereum = ethereum;
+        this.address = address;
+        this.contract = new CallTransaction.Contract(abi);
+        this.depositFilter = new LogFilter().withContractAddress(address)
```
zilm13: Will it catch calls from another contract calling Deposit contract? If not I'd add comment here

mkalinin: Yes, it will. When `LOG` is called from inside of internal transaction, the address of this transaction callee is being logged.

ethereumj-core/src/main/java/org/ethereum/sharding/domain/Validator.java
mkalinin: Nice thing 👍 Will do

ethereumj-core/src/main/java/org/ethereum/db/BlockStore.java
mkalinin: Maybe it's just better to return blocks in direct order?

ethereumj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
mkalinin: The latter may happen when `from` block is not an ancestor of `to` block. The former... I just couldn't make a look up into a list without being sure that it's not empty.

ethereumj-core/src/main/java/org/ethereum/sharding/ShardingEntry.java
mkalinin: What do you mean? That we _will_ loose user config passed this way?

ethereumj-core/src/main/java/org/ethereum/sharding/config/BeaconConfig.java
mkalinin: Good note. I am not fond of Spring DI due to it's unpredictable behaviour in some cases.
Reading from `@Import` description:
> `@Bean` definitions declared in imported `@Configuration` classes should be accessed by using `@Autowired` injection. Either the bean itself can be autowired, or the configuration class instance declaring the bean can be autowired. The latter approach allows for explicit, IDE-friendly navigation between `@Configuration` class methods.

This standing doesn't literally say that beans autowired in configuration classes are always set before their first usage. But it could be concluded that way. Let's keep it this way unless something is broken.

ethereumj-core/src/main/java/org/ethereum/db/BlockStore.java
zilm13: If you don't need it in backwards order later, it's ok

ethereumj-core/src/main/java/org/ethereum/sharding/ShardingEntry.java
zilm13: You can pass only one config with `ethereumj.conf.res`, so in such case if something is passed, it will be overwritten. It will be cool if we could pass several configs. Not for sharding specifically of course, generally.

ethereumj-core/src/main/java/org/ethereum/sharding/contract/DepositContract.java
zilm13: Cool!

mkalinin:
Covered with tests, tested, fixed, polished and ready for merge after another round of review.
@zilm13 please, take a look

ethereumj-core/src/main/java/org/ethereum/sharding/contract/DepositContract.java
```diff
@@ -0,0 +1,197 @@
...
+import static org.ethereum.util.ByteUtil.longToBytes;
+import static org.ethereum.util.blockchain.EtherUtil.convert;
+
+/**
+ * Delivers an API of validator registration contract.
+ *
+ * <p>
+ *     Main endpoints are:
+ *     <ul>
+ *         <li> {@link #deposit(byte[], long, byte[], byte[], DepositAuthority)}
```
zilm13: I'd add brief description to every endpoint here

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/DepositAuthority.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public interface DepositAuthority {
+
+    /**
+     * Signs given transaction
+     */
+    void sign(Transaction tx);
+
+    /**
+     * Returns and address of held authority
```
zilm13: typo

ethereumj-core/src/main/java/org/ethereum/sharding/manager/ShardingWorldManager.java
```diff
@@ -0,0 +1,128 @@
...
+    private void initDepositContract() {
+        if (getBlockchain().getBestBlock().isGenesis()) {
+            Block best = getBlockchain().getBestBlock();
+
+            // build and execute deploy transaction
+            BigInteger nonce = getRepository().getNonce(DEPLOY_AUTHORITY.getAddress());
+            Integer chainId = config.getBlockchainConfig().getConfigForBlock(0).getChainId();
+            Transaction tx = new Transaction(nonce.toByteArray(), longToBytes(DEPLOY_GAS_PRICE), longToBytes(DEPLOY_GAS_LIMIT),
+                    null, BigInteger.ZERO.toByteArray(), contractConfig.getBin(), chainId) {
+                @Override
```
zilm13: Any alternatives to this kind of hack?
I guess you want it to be catched by logfilter, yeah

ethereumj-core/src/test/java/org/ethereum/sharding/ShardingTestHelper.java
```diff
@@ -0,0 +1,192 @@
...
+        DbFlushManager dbFlushManager = sharding.flushManager = new DbFlushManager(systemProperties, Collections.emptySet(),
+                new WriteCache.BytesKey<>(standaloneBlockchain.getStateDS(), WriteCache.CacheType.SIMPLE)) {
+            @Override
+            public synchronized Future<Boolean> flush() {
+                CompletableFuture<Boolean> ret = new CompletableFuture<>();
+                ret.complete(true);
+                return ret;
+            }
+        };
+        Ethereum ethereum = sharding.ethereum = new EthereumImpl(systemProperties, standaloneBlockchain.getListener()) {
```
zilm13: I don't understand, why it's not executed on StandaloneBlockchain and we need to override it here?

ethereumj-core/src/main/java/org/ethereum/sharding/manager/ShardingWorldManager.java
mkalinin: This is a hack indeed.
It was not enough to just save the code of contract to predefined address. It also requires to run the code which does initial set up of the contract. Thus it needs to be executed as a regular tx but also stored to some certain address and I just haven't found anything less intrusive than address overriding.

ethereumj-core/src/test/java/org/ethereum/sharding/ShardingTestHelper.java
mkalinin: `StandaloneBlockchain` doesn't aggregate `Ethereum` facade, actually `StandaloneBlockchain` is a kind of `Ethereum` facade in terms of API it provides. Since entities that are tested here are built on top of `Ethereum` I decided to override its methods rather than introduce an abstraction that allows to use either `Ethereum` or `StandaloneBlockchain` depending on the case

ethereumj-core/src/main/java/org/ethereum/sharding/manager/ShardingWorldManager.java
zilm13: Ok, I got it.

ethereumj-core/src/test/java/org/ethereum/sharding/ShardingTestHelper.java
zilm13: So as we could need it further not only for sharding, could you create an issue and link to this override

mkalinin: Created a ticket https://github.com/ethereum/ethereumj/issues/1139

PR #1147, opened 2018-08-02
Update vm recursion error logs and defaults to match current env

zilm13:
@mkalinin please, take a look


zilm13:
@mkalinin how about this

Issue #1138, opened 2018-08-01
Improve EthereumListener


mkalinin:
Currently [CompositeEthereumListener](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/listener/CompositeEthereumListener.java#L78) invokes _all_ defined listeners on _any_ occurred event which is pretty inefficient. It also impacts on memory consumption. 
Possible fix: drive listeners in a pub/sub fashion. It would also nice to have a possibility to define one-off listeners that are detached when a certain event has been caught.

Recap:
- use pub/sub pattern to invoke listener only if it has a subscription to a certain event
- provide a kind of `evict()` method that removes listener when current invocation is finished
- make listener for one-off events like `onSyncDone` remove itself in automatic fashion
- get rid of [trace()](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java#L100) and its usage all over the codebase

Example of listener with evict:
```
ethereum.addListener(new EthereumListenerAdapter() {
    @Override
    public void onPendingTransactionUpdate(TransactionReceipt txReceipt, PendingTransactionState state, Block block) {
        if (FastByteComparisons.equal(depositTxHash, txReceipt.getTransaction().getHash()) &&
                state == PendingTransactionState.INCLUDED) {
            depositFuture.complete(txReceipt);

            evict();
        }
    }
});
```

PR #1152, opened 2018-08-10
Improve EthereumListener

eugene-shevchenko:
This PR covers issue [#1138](https://github.com/ethereum/ethereumj/issues/1138).

Adds Publisher abstraction that implements pub/sub model to deliver events from publisher to subscriber.
Replaces CompositeEthereumListener / EthereumListenerAdaptor / EthereumListener using with new event publishing mechanism.
Keeps backward compatibility with old event handling model.


ethereumj-core/src/main/java/org/ethereum/publish/exprimental/EventBus.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.Consumer;
+
+import static java.util.Objects.isNull;
+
+public class EventBus {
```
mkalinin: Maybe it is better to move `experimental` stuff to a separate branch?

ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
```diff
@@ -222,9 +225,14 @@ private void resetDataSource(Source source) {
...
-    public CompositeEthereumListener ethereumListener() {
-        return new CompositeEthereumListener();
+    @Bean
+    public Publisher publisher(EventDispatchThread eventDispatchThread) {
+        return new Publisher(eventDispatchThread);
+    }
+
+    @Bean
+    public EthereumListener ethereumListener(Publisher publisher) {
+        return publisher.asListener();
```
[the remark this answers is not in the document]
eugene-shevchenko: 1. Do you mean to move backward compatibility logic from Publisher to external adapter class, to keep Publisher code clean?
2. For each callback from EthereumListener exists similar Event type, that could be fired and caught.

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
```diff
@@ -47,20 +52,11 @@
...
-import java.util.ConcurrentModificationException;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
```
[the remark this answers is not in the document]
eugene-shevchenko: Ok, I'll move back direct class import. It's just IDEA automatic import optimizer.

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
```diff
@@ -455,8 +454,10 @@ public synchronized ImportResult tryToConnect(final Block block) {
         }
 
         if (ret.isSuccessful()) {
-            listener.onBlock(summary, ret == IMPORTED_BEST);
-            listener.trace(String.format("Block chain size: [ %d ]", this.getSize()));
+            publisher
+                    .publish(new BlockAddedEvent(summary))
+                    .publish(new BestBlockAddedEvent(summary, ret == IMPORTED_BEST))
```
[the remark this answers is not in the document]
eugene-shevchenko: It was added for backward compatibility. The old code had default callback implementation with data proxing. So to fire such client code in a new fashion we need generate both events. 
I agree, that we should leave only one event, when we'll completely remove EthereumListener.

ethereumj-core/src/main/java/org/ethereum/publish/exprimental/EventBus.java
eugene-shevchenko: I'll remove this package completely.

ethereumj-core/src/main/java/org/ethereum/publish/event/NoConnectionsEvent.java
```diff
@@ -0,0 +1,4 @@
+package org.ethereum.publish.event;
+
+public class NoConnectionsEvent extends SignalEvent {
```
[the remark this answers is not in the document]
eugene-shevchenko: Yep, it was created for correspond EthereumListener callback, before all invocations were migrated. I'll remove it.

ethereumj-core/src/main/java/org/ethereum/publish/event/BestBlockAddedEvent.java
```diff
@@ -0,0 +1,28 @@
+package org.ethereum.publish.event;
+
+import org.ethereum.core.BlockSummary;
+
+public class BestBlockAddedEvent extends Event<BestBlockAddedEvent.Data> {
```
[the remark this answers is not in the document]
eugene-shevchenko: Can't argue with that :) I had the same idea too, cause sometimes such naming lead to worst enterprise apps class names. Will rename all events.

ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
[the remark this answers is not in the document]
mkalinin: 1. I mean that it makes sense to keep both old listener which stays unchanged and a publisher. Listener will be marked as deprecated and publisher won't include already deprecated parts like `onPendingTransactionReceived()` method

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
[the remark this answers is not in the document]
mkalinin: Actually, it depends on configuration. You may disable wildcards in idea.

ethereumj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
```diff
@@ -104,20 +109,25 @@
...
         this.config = config;
         System.out.println();
-        this.compositeEthereumListener.addListener(gasPriceTracker);
         gLogger.info("EthereumJ node started: enode://" + toHexString(config.nodeId()) + "@" + config.externalIp() + ":" + config.listenPort());
     }
 
+    @PostConstruct
+    public void init() {
+        publisher.subscribe(to(BlockAdded.class, data -> {
+            data.getBlockSummary().getBlock().getTransactionsList().forEach(gasPriceTracker::onTransaction);
```
zilm13: this logic should be somewhere in gasPriceTracker instead

ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
```diff
@@ -27,18 +27,22 @@
 import java.util.List;
 
 /**
+ * Uses expensive enough logic for delivery events to client.<br>
+ * Will be removed in future releases.
+ *
  * @author Roman Mandeleil
  * @since 27.07.2014
+ * @deprecated use {@link org.ethereum.publish.Publisher} instead.
```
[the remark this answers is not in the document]
zilm13: We'll keep the interface just for these enums? Maybe just copy it somewhere right now and recommend using the copy?

ethereumj-core/src/main/java/org/ethereum/facade/Ethereum.java
```diff
@@ -53,8 +56,27 @@
...
+    /**
+     * Adds subscription for specific event.
+     *
+     * @param subscription
+     * @return publisher instance to support fluent API.
+     * @see Subscription
+     * @see Publisher
+     * @see Event
+     */
+    Publisher subscribe(Subscription subscription);
```
mkalinin: As a user I'd rather expect `subscribe(Class type, Consumer handler);` which is clearer. But we have to use generics though to make it really convenient:
`<T> Publisher subscribe(Class<? extends Event<T>> type, Consumer<T> handler);`

ethereumj-core/src/main/java/org/ethereum/listener/BackwardCompatibilityEthereumListenerProxy.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Override
+    public void trace(String output) {
+        compositeListener.trace(output);
+    }
+
+    @Override
+    public void onNodeDiscovered(Node node) {
+        compositeListener.onNodeDiscovered(node);
+        publisher.publish(Events.onNodeDiscovered(node));
```
mkalinin: Maybe use static import for `Events`?

ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
```diff
@@ -221,9 +235,14 @@ private void resetDataSource(Source source) {
...
-    @Bean(name = "EthereumListener")
-    public CompositeEthereumListener ethereumListener() {
-        return new CompositeEthereumListener();
+    @Bean
+    public Publisher publisher(EventDispatchThread eventDispatchThread) {
+        return new Publisher(eventDispatchThread);
+    }
+
+    @Bean
+    public CompositeEthereumListener compositeEthereumListener(EventDispatchThread eventDispatchThread) {
```
mkalinin: Why don't just instantiate `BackwardCompatibilityEthereumListenerProxy` here?

ethereumj-core/src/main/java/org/ethereum/listener/CompositeEthereumListener.java
```diff
@@ -50,10 +51,14 @@ public String toString() {
...
 
-    @Autowired
-    EventDispatchThread eventDispatchThread = EventDispatchThread.getDefault();
-    
-    protected List<EthereumListener> listeners = new CopyOnWriteArrayList<>();
+    private final Executor executor;
+    private final List<EthereumListener> listeners;
+
+
+    public CompositeEthereumListener(Executor executor) {
```
mkalinin: Why do we have anything to update in deprecated class?

ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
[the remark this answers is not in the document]
mkalinin: +1 for the copy

ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
```diff
@@ -150,4 +160,86 @@ default void onBlock(BlockSummary blockSummary, boolean best) {
     void onTransactionExecuted(TransactionExecutionSummary summary);
 
     void onPeerAddedToSyncPool(Channel peer);
+
+    EthereumListener STUB = new EthereumListener() {
```
mkalinin: I don't think this either necessary. Let it use old deprecated `EthereumListenerAdapter`

ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
```diff
@@ -38,7 +38,7 @@
  * (if {@link #getPercentileShare()} is not overridden) of the latest transactions were
  * executed at this or lower price.
  */
-public class RecommendedGasPriceTracker extends EthereumListenerAdapter {
+public class RecommendedGasPriceTracker {
```
mkalinin: This change breaks backward compatibility. We also need to introduce a mechanism for easy subscribing to the new publisher.

ethereumj-core/src/main/java/org/ethereum/publish/event/BlockAdded.java
```diff
@@ -0,0 +1,28 @@
...
+
+        public boolean isBest() {
+            return best;
+        }
+    }
+
+    public BlockAdded(BlockSummary blockSummary, boolean best) {
+        super(new Data(blockSummary, best));
+    }
+}
```
mkalinin: new line

ethereumj-core/src/main/java/org/ethereum/publish/LifeCycleSubscription.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.function.BiConsumer;
+
+/**
+ * Abstraction that holds together event type, event processor (a.k.a consumer) and optional conditions that resolves
+ * event processing and auto-unsubscribe.
+ *
+ * @author Eugene Shevchenko
+ */
+public class LifeCycleSubscription<E extends Event<D>, D> extends Subscription<E, D> {
```
mkalinin: I bet we don't need yet another abstraction here. Introducing it makes an API less obvious for users. I am more about to have two constructors for `Subscription`, one with `BiConsumer`, another one with `Consumer` that is extended to `BiConsumer` during construction.

ethereumj-core/src/main/java/org/ethereum/publish/LifeCycleSubscription.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static class LifeCycle {
```
mkalinin: Let's make it simple. Each subscription has a lifecycle which just holds flags that user is gonna change. We may just add a check like `lifeCycle.unsubscribed || ...` to `needUnsubscibeAfter` method.

ethereumj-core/src/main/java/org/ethereum/listener/BackwardCompatibilityEthereumListenerProxy.java
eugene-shevchenko: Nope, we have several methods with the same signature in both classes. That's why we should use full name in this case. In addition it's temporary class, so it's not critical I suppose.

ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
eugene-shevchenko: Does anyone use this class except us? I thought that this class for our internal logic only.

ethereumj-core/src/main/java/org/ethereum/listener/CompositeEthereumListener.java
eugene-shevchenko: Firstly field injection is bad practice cause it hides component dependencies. But in this case it's not critical because class marked as deprecated.
I moved bean definition to java config to highlight that old and new components use the same `Executor`. And then both `Publisher` and deprecated `CompositeEthereumListener` injects into `BackwardCompatibilityEthereumListenerProxy`. IMHO it's more obvious and helps avoid dependency mess. 

This also applies to the previous comment.

ethereumj-core/src/main/java/org/ethereum/facade/Ethereum.java
eugene-shevchenko: I'll add two overloaded methods: 
 * `<T> Publisher subscribe(Class<? extends Event<T>> type, Consumer<T> handler);`
 * `<T> Publisher subscribe(Class<? extends Event<T>> type, BiConsumer<T, Subscription.LifeCycle> handler);
`
but we need current method too, cause sometimes we are in want of `Subscription` instance.

ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
```diff
@@ -290,14 +300,14 @@ public DbFlushManager dbFlushManager() {
     }
 
     @Bean
-    public BlockHeaderValidator headerValidator() {
+    public BlockHeaderValidator headerValidator(SystemProperties systemProperties, Publisher publisher) {
```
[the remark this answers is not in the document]
eugene-shevchenko: The method uses only in java config, and as I mentioned earlier such bean definition hides component dependencies. With no arguments option I should invoke three factory methods in method's body to get necessary dependencies. But if you think that no-args option is better I can revert method's signature.   

PS: It's sad that java config which should configure `ApplicationContext` only, we use like object factory.

ethereumj-core/src/main/java/org/ethereum/publish/LifeCycleSubscription.java
eugene-shevchenko: Ok, I'll move this functionality to `Subscription`

ethereumj-core/src/main/java/org/ethereum/publish/LifeCycleSubscription.java
eugene-shevchenko: Ok, I'll move this functionality to `Subscription`

ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
eugene-shevchenko: Ok, I'll revert this change.

ethereumj-core/src/main/java/org/ethereum/listener/EthereumListener.java
eugene-shevchenko: Ok, I'll replace it with `EthereumListenerAdapter`

ethereumj-core/src/main/java/org/ethereum/listener/RecommendedGasPriceTracker.java
eugene-shevchenko: By the way we use this class nowhere =)

ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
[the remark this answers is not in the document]
zilm13: I think it's not a bad thing if we want to get rid of Spring one day.
It's more deterministic way.

ethereumj-core/src/main/java/org/ethereum/core/BlockchainImpl.java
```diff
@@ -209,11 +223,6 @@ public BlockchainImpl withAdminInfo(AdminInfo adminInfo) {
         return this;
     }
 
-    public BlockchainImpl withEthereumListener(EthereumListener listener) {
```
zilm13: Listener doesn't look the thing which is required for running `BlockchainImpl`. Also old constructor could be used by somebody. I'd prefer to keep both old constructor (and init listener there with `EthereumListener.EMPTY`) and `withEthereumListener` method.

ethereumj-core/src/main/java/org/ethereum/manager/WorldManager.java
```diff
@@ -121,9 +124,25 @@ private void init() {
...
+
+    /**
+     * @param listener
+     * @deprecated use #subscription instead.
+     */
+    @Deprecated
     public void addListener(EthereumListener listener) {
         logger.info("Ethereum listener added");
-        ((CompositeEthereumListener) this.listener).addListener(listener);
+        ((BackwardCompatibilityEthereumListenerProxy) listener).getCompositeListener().addListener(listener);
```
zilm13: If user have used our listener in most common way (`ethereum.addListener(new EthereumListenerAdapter(){})`) he will get strange exception here, I guess.

ethereumj-core/src/main/java/org/ethereum/publish/event/Events.java
```diff
@@ -0,0 +1,95 @@
...
+import org.ethereum.net.message.Message;
+import org.ethereum.net.p2p.HelloMessage;
+import org.ethereum.net.rlpx.Node;
+import org.ethereum.net.server.Channel;
+import org.ethereum.publish.event.message.EthStatusUpdated;
+import org.ethereum.publish.event.message.MessageReceived;
+import org.ethereum.publish.event.message.MessageSent;
+import org.ethereum.publish.event.message.PeerHandshaked;
+
+public final class Events {
```
zilm13: Does this class made for internal purposes of compatibility?

ethereumj-core/src/main/java/org/ethereum/facade/Ethereum.java
zilm13: We should add some handy links where user could see available type of Events. Current user entrance could stuck without samples.

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/EthHandler.java
```diff
@@ -99,6 +104,10 @@ public void channelRead0(final ChannelHandlerContext ctx, EthMessage msg) throws
         msgQueue.receivedMessage(msg);
     }
 
+    public Publisher getPublisher() {
+        return ((BackwardCompatibilityEthereumListenerProxy) ethereumListener).getPublisher();
```
zilm13: I think `BackwardCompatibilityEthereumListenerProxy` should be localized somehow and don't spread to such classes.

ethereumj-core/src/main/java/org/ethereum/facade/Ethereum.java
eugene-shevchenko: Each of samples migrated to the new feature, so I guess it's not problem. Also we have utility class `Events` for convenience instantiation all of supported events. I'd rather actualize `Publisher` javadoc to refer user to all necessary classes.

ethereumj-core/src/main/java/org/ethereum/manager/WorldManager.java
eugene-shevchenko: Why do you think so ? This code fully support old logic.

ethereumj-core/src/main/java/org/ethereum/publish/event/Events.java
eugene-shevchenko: Nope, this class added for using convenience. For instance using 
`publisher.publish(onSyncDone(SOME_STATUS))` more readable than `publisher.publish(new SyncDone(SOME_STATUS))`.

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/EthHandler.java
eugene-shevchenko: The main idea of `BackwardCompatibilityEthereumListenerProxy` is to encapsulate new and old publishing mechanisms under` EthereumListener` interface, and proxy old messages to both of them with / without transformation. We should use this component everywhere to reduce the number of changes and support both mechanisms at the same time.
`BackwardCompatibilityEthereumListenerProxy` will simplify removing deprecated logic in the future. And when the time comes we'll remove `BackwardCompatibilityEthereumListenerProxy` and all` EthereumListener` related stuff and leave only `Publisher` invokes.

ethereumj-core/src/main/java/org/ethereum/facade/Ethereum.java
zilm13: Before this update user was adding listener by extending `EthereumListenerAdapter`, he got the view of all available listening options in one click. Since this update he should write `.subscribe(` and what's next? Samples are not the answer, not everyone can find them and you will not go to samples every time you need to add listener, Subscription class looks difficult from the first view, events are different classes in some package somewhere, not listed in one place. So the idea is that main entrance should look handy and simplify user's usage. I'm not sure what's the best way to do it but current entrance could stuck most of users.

ethereumj-core/src/main/java/org/ethereum/publish/event/Events.java
zilm13: Got it

ethereumj-core/src/main/java/org/ethereum/samples/EventListenerSample.java
```diff
@@ -287,13 +283,8 @@ private void waitForEther(byte[] address, BigInteger requiredBalance) throws Int
...
     private void deployContractAndTest() throws Exception {
-        ethereum.addListener(new EthereumListenerAdapter() {
-            // when block arrives look for our included transactions
-            @Override
-            public void onBlock(Block block, List<TransactionReceipt> receipts) {
-                EventListenerSample.this.onBlock(block, receipts);
-            }
-        });
+        // when block arrives look for our included transactions
+        ethereum.subscribe(to(BlockAdded.class, this::onBlock));
```
zilm13: I see old method usage in `private void replayOnly() throws Exception`
`ethereum.addListener(blockReplay);`

ethereumj-core/src/main/java/org/ethereum/manager/WorldManager.java
zilm13: Users have code like
`ethereum.addListener(new EthereumListenerAdapter());`
after that it goes:
-> 
`worldManager.addListener(listener)`
->
`((BackwardCompatibilityEthereumListenerProxy) EthereumListenerAdapter instance)`

ethereumj-core/src/main/java/org/ethereum/net/eth/handler/EthHandler.java
zilm13: I got this idea as we are moving everything inside our application to the new subscribe-publish model, while leaving common user interfaces and marking them as deprecated. So any backward compatibility gateways should be localized to old public user interfaces.

ethereumj-core/src/main/java/org/ethereum/listener/BackwardCompatibilityEthereumListenerProxy.java
```diff
@@ -0,0 +1,187 @@
...
+    public Publisher getPublisher() {
+        return publisher;
+    }
+
+    public void addListener(EthereumListener listener) {
+        this.compositeListener.addListener(listener);
+    }
+
+    @Override
+    public void trace(String output) {
```
zilm13: why this kind of event was not added to publisher-subscriber implementation?

ethereumj-core/src/main/java/org/ethereum/facade/Ethereum.java
```diff
@@ -53,8 +58,37 @@
...
+
+    /**
+     * More convenient version of {@link #subscribe(Subscription)}
+     */
+    <T> Publisher subscribe(Class<? extends Event<T>> type, Consumer<T> handler);
+
+    /**
+     * More convenient version of {@link #subscribe(Subscription)}
+     */
+    <T> Publisher subscribe(Class<? extends Event<T>> type, BiConsumer<T, Subscription.LifeCycle> handler);
```
mkalinin: I think `Publisher` class should also have such shortcuts

mkalinin: To increase UX we may create shortcuts for event classes and put them together into one place.
For example, use `Event` or create new `EventType` class and add a list of event types definition with readable names:
`public static final Class<? extends Event<BlockAdded.Data>> BLOCK_ADDED = BlockAdded.class;`
Then point this type aggregation class in javadocs. It would work like an event picker then and make user choice faster, user will have to type `Event.` to get a list of available options

ethereumj-core/src/main/java/org/ethereum/listener/BlockReplayer.java
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Class capable of replaying stored blocks prior to 'going online' and
+ * notifying on newly imported blocks
+ * <p>
+ * For example of usage, look at {@link org.ethereum.samples.BlockReplaySample}
+ * <p>
+ * Created by Eugene Shevchenko on 07.10.2018.
+ */
+public class BlockReplayer {
```
mkalinin: Should we mark its predecessor `BlockReplay` class as a deprecated one?

ethereumj-core/src/main/java/org/ethereum/publish/Subscription.java
```diff
@@ -0,0 +1,169 @@
...
+    private final static Logger log = LoggerFactory.getLogger("event");
+
+    public static class LifeCycle {
+        private final Subscription subscription;
+
+        private LifeCycle(Subscription subscription) {
+            this.subscription = subscription;
+        }
+
+        public void unsubscribe() {
```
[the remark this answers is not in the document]
mkalinin: still not added :)

ethereumj-core/src/main/java/org/ethereum/db/DbFlushManager.java
```diff
@@ -44,7 +51,7 @@
 
     List<AbstractCachedSource<byte[], ?>> writeCaches = new CopyOnWriteArrayList<>();
     List<Source<byte[], ?>> sources = new CopyOnWriteArrayList<>();
-    Set<DbSource> dbSources = new HashSet<>();
+    Set<DbSource> dbSources ;
```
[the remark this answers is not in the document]
eugene-shevchenko: I'll revert this change

ethereumj-core/src/main/java/org/ethereum/listener/BlockReplayer.java
eugene-shevchenko: yep, I'll add comment.

PR #1155, opened 2018-08-16
Fix the case: when switching off the sync after fastsync

Nashatyrev:
I did fastsync, then switched off the sync completely and my db was cleaned up

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
+++ b/ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
@@ -195,13 +195,14 @@ public class CommonConfig {
     }
 
     public void fastSyncCleanUp() {
+        if (!systemProperties().isSyncEnabled()) return;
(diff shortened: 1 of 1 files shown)


ethereumj-core/src/main/java/org/ethereum/config/CommonConfig.java
```diff
@@ -201,7 +201,7 @@ public void fastSyncCleanUp() {
 
         EthereumListener.SyncState syncStage = EthereumListener.SyncState.values()[fastsyncStageBytes[0]];
 
-        if (!systemProperties().isFastSyncEnabled() || syncStage == EthereumListener.SyncState.UNSECURE) {
+        if ((systemProperties().isSyncEnabled() && !systemProperties().isFastSyncEnabled()) || syncStage == EthereumListener.SyncState.UNSECURE) {
```
mkalinin: It's better to add this check to the beginning of the method cause whole method call doesn't make sense if sync is OFF

Issue #1157, opened 2018-08-19
NodeKeyCompositor for storage [ BUG ]


ithinker1991:
ethereumj-core/src/main/java/org/ethereum/datasource/NodeKeyCompositor.java
This method
``` java
 private static byte[] composeInner(byte[] key, byte[] addrHash) {

        validateKey(key);

        byte[] derivative = new byte[key.length];
        arraycopy(key, 0, derivative, 0, PREFIX_BYTES);
        arraycopy(addrHash, 0, derivative, PREFIX_BYTES, PREFIX_BYTES);

        return derivative;
}
```

``` java
 public static void main(String[] args) {
        byte[] addrhash = Hex.decode("a9539c810cc2e8fa20785bdd78ec36ccb25e1b5be78dbadf6c4e817c6d170bbb");
        byte[] key = Hex.decode("bbbbbb5be78dbadf6c4e817c6d170bbb47e9916f8f6cc4607c5f3819ce98497b");

        byte[] ret = NodeKeyCompositor.compose(key, addrhash);
        System.out.println(Hex.toHexString(ret));

       // output : bbbbbb5be78dbadf6c4e817c6d170bbba9539c810cc2e8fa20785bdd78ec36cc
    }
```

the output is not  match the example 
 * Example: <br/>
 * Contract address hash <i>a9539c810cc2e8fa20785bdd78ec36ccb25e1b5be78dbadf6c4e817c6d170bbb</i> <br/>
 * Key of one of the storage nodes <i>bbbbbb5be78dbadf6c4e817c6d170bbb47e9916f8f6cc4607c5f3819ce98497b</i> <br/>
 * Composed key will be <i>a9539c810cc2e8fa20785bdd78ec36ccbbbbbb5be78dbadf6c4e817c6d170bbb</i> <br/><br/>

mkalinin:
Yep, that was incorrect. Fixed by https://github.com/ethereum/ethereumj/commit/ec3ce8fccb75987f101096c61af137c4f5b17adb. Thank you for the report!

ithinker1991:
emmm...
There seems a another BIG problem. I noticed in solidity, for dynamic type such as  `string`， `bytes`
head of the storage-key are same and tail are different.
So if one field has been encoded to many storage-key. the composed key will be same.

For example
string s = '111111111111111111111111111111111111111111'
encoded result:
key of storage node												 value of storage node													     
[...]

PR #1158, opened 2018-08-20
fix same composed key of contract storage nodes

ithinker1991:
32-bytes composed key should be 
[first 16-bytes of address hash:second 16-bytes of node key] 

There seems a  BIG problem. I noticed in solidity, for dynamic type such as  `string`， `bytes`
first 16-bytes of node-key are same and second 16-bytes of node are different.
So if one field has been encoded to many storage-key. the composed key will be same.

For example
        s1 = "1111111111111111111111111111111111111112222222222222222222222222222222222222222333333333333333333333333333333333333";

encoded result:
```
key of storage node								value of storage node													     
0000000000000000000000000000000000000000000000000000000000000000 00000000000000000000000000000000000000000000000000000000000000e7
290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563 3131313131313131313131313131313131313131313131313131313131313131
290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e564 3131313131313132323232323232323232323232323232323232323232323232
290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e565 3232323232323232323232323232323333333333333333333333333333333333
290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e566 3333333333333333333333333333333333333300000000000000000000000000
```

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/datasource/NodeKeyCompositor.java
+++ b/ethereumj-core/src/main/java/org/ethereum/datasource/NodeKeyCompositor.java
@@ -12,13 +12,14 @@ import static org.ethereum.crypto.HashUtil.sha3;
  * <p>
  *     <b>Input:</b> 32-bytes node key, 20-bytes contract address
  * <br/>
- *     <b>Output:</b> 32-bytes composed key <i>[first 16-bytes of node key : first 16-bytes of address hash]</i>
+ *     <b>Output:</b> 32-bytes composed key <i>[first 16-bytes of address hash : second 16-bytes of node key]</i>
  *
  * <p>
  *     Example: <br/>
  *     Contract address hash <i>a9539c810cc2e8fa20785bdd78ec36ccb25e1b5be78dbadf6c4e817c6d170bbb</i> <br/>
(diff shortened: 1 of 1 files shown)


ithinker1991:
https://github.com/ethereum/ethereumj/issues/1157

mkalinin:
There is a path that solidity uses to address storage variable. It is a 32-byte value that is defined by compiler for each storage variable. Storage key is not the same as storage variable path, it's a sha3 from that path, `key = sha3(path)`. Thus lower 16 bytes of the key reflect whole path value whatever it is.

ithinker1991:
@mkalinin thank you for your reply. But 
I have see `sha3(path)` in class `SecureTrie`. The class `SecureTrie`  is used in many scenarios.
Why uses `SecureTrie` instead of `TrieImp`？

Nashatyrev:
@ithinker1991 because of Ethereum specification.
Originally this was done to avoid malicious Trie manipulation. I.e. you may intentionally add such data to a contract that Trie depth becomes very large which causes significant performance degradation for Trie operations

PR #1160, opened 2018-08-20
Fix discovery exception handling

mkalinin:
Fixes `UDPListener` crash caused by malformed `nodeId` received from malicious peers.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/net/rlpx/discover/MessageHandler.java
+++ b/ethereumj-core/src/main/java/org/ethereum/net/rlpx/discover/MessageHandler.java
@@ -52,7 +52,7 @@ public class MessageHandler extends SimpleChannelInboundHandler<DiscoveryEvent>
         try {
             nodeManager.handleInbound(event);
         } catch (Throwable t) {
-            logger.error("Failed to process incoming message: {}", event.getMessage(), t);
+            logger.info("Failed to process incoming message: {}, caused by: {}", event.getMessage(), t.toString());
         }
     }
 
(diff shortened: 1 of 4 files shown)


zilm13:
I think you should disconnect it in case of error anyway, should you?

mkalinin:
@zilm13 
> I think you should disconnect it in case of error anyway, should you?

UDP is connectionless protocol, thus there is no session that can be closed.

mkalinin:
Thanks a lot to @JMdoubleU for pointing out to that problem!

ethereumj-core/src/main/java/org/ethereum/net/rlpx/discover/MessageHandler.java
```diff
@@ -49,7 +49,11 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
 
     @Override
     public void channelRead0(ChannelHandlerContext ctx, DiscoveryEvent event) throws Exception {
-        nodeManager.handleInbound(event);
+        try {
+            nodeManager.handleInbound(event);
+        } catch (Throwable t) {
+            logger.error("Failed to process incoming message: {}", event.getMessage(), t);
```
Nashatyrev: I would change log level to `WARN` or even `INFO`, since we can receive any garbage and this is not our error. 
I'd also output `t.toString()` instead of `t` to not flood logs with traces in case of invalid messages 

Ideally we can also try to differentiate the message handle problems onto `MalformedMessageException` and others. The first one is not error, the second one should be assumed as our error

zilm13:
@mkalinin 
> UDP is connectionless protocol, thus there is no session that can be closed.
oops, forgot that discovery works on UDP

ethereumj-core/src/main/java/org/ethereum/net/rlpx/discover/MessageHandler.java
mkalinin: Decided to reduce logging level for the beginning. `MalformedMessageException` requires much more changes in the codebase.

PR #1162, opened 2018-08-20
Feature/eip 1052

zilm13:
Do not merge, `GitHubBlockStateTest` tests are not passed.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/db/RepositoryImpl.java
+++ b/ethereumj-core/src/main/java/org/ethereum/db/RepositoryImpl.java
@@ -135,7 +135,7 @@ public class RepositoryImpl implements Repository, org.ethereum.facade.Repositor
     @Override
     public synchronized byte[] getCode(byte[] addr) {
         byte[] codeHash = getCodeHash(addr);
-        return FastByteComparisons.equal(codeHash, HashUtil.EMPTY_DATA_HASH) ?
+        return codeHash == null || FastByteComparisons.equal(codeHash, HashUtil.EMPTY_DATA_HASH) ?
                 ByteUtil.EMPTY_BYTE_ARRAY : codeCache.get(codeKey(codeHash, addr));
     }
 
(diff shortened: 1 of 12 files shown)


ethereumj-core/src/main/java/org/ethereum/db/RepositoryImpl.java
```diff
@@ -147,7 +147,7 @@ public synchronized void saveCode(byte[] addr, byte[] code) {
     @Override
     public byte[] getCodeHash(byte[] addr) {
         AccountState accountState = getAccountState(addr);
-        return accountState != null ? accountState.getCodeHash() : HashUtil.EMPTY_DATA_HASH;
+        return accountState != null ? accountState.getCodeHash() : null;
```
mkalinin: Are you sure that this change won't break anything else than `getCode`?

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -871,6 +880,23 @@ else if (oldValue != null && newValue.isZero()) {
                     program.step();
                 }
                 break;
+                case EXTCODEHASH: {
+                    DataWord address = program.stackPop();
+                    PrecompiledContracts.PrecompiledContract contract =
+                            PrecompiledContracts.getContractForAddress(address, blockchainConfig);
+
+                    byte[] codeHash;
+                    // The EXTCODEHASH of an precompiled contract is either c5d246... or 0
```
mkalinin: I think it means that if precompile acc is non-existent the hash is 0 otherwise the hash is an empty data hash.

mkalinin: Hence, precompile contract doesn't need a special case to be processed

ethereumj-core/src/main/java/org/ethereum/db/RepositoryImpl.java
zilm13: @mkalinin checked it with extra attention, it shouldn't break anything for us

PR #1164, opened 2018-08-22
Only process neighbors when they were requested

JMdoubleU:


mkalinin:
@JMdoubleU thanks for that update! Looks like a reasonable thing. As to me, it's ready for merge

PR #1167, opened 2018-08-23
Feature/eip 145

zilm13:
@mkalinin please, take a look!
Also I'd separate VMTest to several smaller tests, what do you think?


zilm13:
Oops, VMTest is running with FrontierConfig. I'll fix it.

ethereumj-core/src/main/java/org/ethereum/vm/DataWord.java
```diff
@@ -373,6 +362,36 @@ public DataWord mulmod(DataWord word1, DataWord word2) {
         return new DataWord(ByteUtil.copyToArray(result.and(MAX_VALUE)));
     }
 
+    /**
+     * Shift left, both this and input arg are treated as unsigned
+     * @param arg
+     * @return this << arg
+     */
+    public DataWord shiftLeft(DataWord arg) {
+        BigInteger result = value().shiftLeft(arg.value().intValue());
```
mkalinin: It's better to use `intValueSafe()` here to avoid undesirable overflow

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -671,6 +674,61 @@ else if (oldValue != null && newValue.isZero()) {
                     program.step();
                 }
                 break;
+                case SHL: {
+                    DataWord word1 = program.stackPop();
+                    DataWord word2 = program.stackPop();
+                    final DataWord result;
+                    if (word1.value().compareTo(BigInteger.valueOf(256)) < 0) {
```
mkalinin: This and following checks (in other similar instructions) is better to move to `DataWord` class cause they also make sense for `DataWord` itself

PR #1169, opened 2018-09-04
Adding ability to uses lists of configs, both with resources and files

zilm13:
We need this feature badly, for example, for Harmony.


ethereumj-core/src/test/java/org/ethereum/config/SystemPropertiesTest.java
```diff
@@ -518,6 +518,30 @@ public void testDump() {
...
+    @Test
+    public void testMergeConfigs2() {
+        String firstConfig = "peer.listen.port=30123";
+        SystemProperties props = new SystemProperties();
+        Config config = props.mergeConfigs(firstConfig, ConfigFactory::parseString);
+        assertEquals(30123, config.getInt("peer.listen.port"));
+    }
+
+    @Test
+    public void testMergeConfigs3() {
```
mkalinin: It lacks of test case with different properties that are eventually merged into single configuration.

PR #1173, opened 2018-09-06
Feature/eip 1283

zilm13:
Have not found any github tests for this, only unit provided

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/Repository.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Repository.java
@@ -235,4 +235,6 @@ public interface Repository extends org.ethereum.facade.Repository{
                      HashMap<ByteArrayWrapper, ContractDetails> cacheDetails);
 
     Repository getSnapshotTo(byte[] root);
+
+    Repository clone();
 }
--- a/ethereumj-core/src/main/java/org/ethereum/core/Repository.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Repository.java
@@ -236,5 +236,9 @@ public interface Repository extends org.ethereum.facade.Repository{
 
     Repository getSnapshotTo(byte[] root);
 
+    /**
+     * Clones repository so changes made to this repository are
+     * not reflected in its clone. 
+     */
     Repository clone();
 }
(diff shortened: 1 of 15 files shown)


ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -136,6 +137,7 @@ public Program(byte[] codeHash, byte[] ops, ProgramInvoke programInvoke, Transac
         traceListener = new ProgramTraceListener(config.vmTrace());
         this.memory = setupProgramListener(new Memory());
         this.stack = setupProgramListener(new Stack());
+        this.originalRepo = programInvoke.getRepository().getSnapshotTo(programInvoke.getRepository().getRoot());
```
zilm13: Ooops, we could have not only `RepositoryImpl` here at least in tests.
Any ideas how can we get copy of Repo here?

ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -789,10 +791,29 @@ private int getReturnDataBufferSizeI() {
                 Arrays.copyOfRange(returnDataBuffer, off.intValueSafe(), off.intValueSafe() + size.intValueSafe());
     }
 
+    @Deprecated
```
mkalinin: I'd not deprecate `storageLoad` cause `SLOAD` op should keep using it. `getCurrentValue` and `storageLoad` are different operations in terms of semantics, I'd decouple them from each other.

ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -789,10 +791,29 @@ private int getReturnDataBufferSizeI() {
...
 
+    /**
+     * @return current Storage data for key
+     */
+    public DataWord getCurrentValue(DataWord key) {
+        return getStorage().getStorageValue(getOwnerAddress().getLast20Bytes(), key);
+    }
+
+    /*
+     * @return Storage data at the beginning of Program execution
```
mkalinin: Let's extend description with a reference to the EIP-1283

ethereumj-core/src/main/java/org/ethereum/core/Repository.java
```diff
@@ -235,4 +235,6 @@ void loadAccount(byte[] addr, HashMap<ByteArrayWrapper, AccountState> cacheAccou
                      HashMap<ByteArrayWrapper, ContractDetails> cacheDetails);
 
     Repository getSnapshotTo(byte[] root);
+
+    Repository clone();
```
mkalinin: Let's add a javadoc to highlight semantics of this call.

PR #1178, opened 2018-09-13
EIP-1014 Skinny CREATE2

zilm13:
Few things not yet finished, do not merge until resolved:
- 1 of 111 test failed both in StateTests and BlockStateTests (same), still finding out the reason
- Do we need to extend`Transaction#getContractAddress()` for `CREATE2`? It's not straightforward, code need to be partially executed for this.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
+++ b/ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
@@ -413,7 +413,7 @@ public class Program {
         byte[] senderAddress = this.getOwnerAddress().getLast20Bytes();
         BigInteger endowment = value.value();
         if (!verifyCall(senderAddress, endowment))
-            return;;
+            return;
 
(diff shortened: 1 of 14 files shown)


ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -401,25 +401,76 @@ public Repository getStorage() {
...
+     * @param memStart      Code memory offset
+     * @param memSize       Code memory size
+     */
     public void createContract(DataWord value, DataWord memStart, DataWord memSize) {
         returnDataBuffer = null; // reset return buffer right before the call
 
+        byte[] senderAddress = this.getOwnerAddress().getLast20Bytes();
+        BigInteger endowment = value.value();
+        if (!verifyCall(senderAddress, endowment))
+            return;;
```
mkalinin: Double `;` 😀

zilm13:
@mkalinin I think we could merge it and move the tests later alone

PR #1179, opened 2018-09-14
VMHook redesign

eugene-shevchenko:
Removes VMHookFactory and VMHook static method injection.
Adds new mechanism VMHook initialization via Spring context.
To add custom VMHook implementation it just need define Spring component within ethereumj context.


ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1408,42 +1400,17 @@ public void play(Program program) {
...
-        }
-
-        if (globalVmHook != null) {
-            try {
-               action.accept(globalVmHook, program);
-            } catch (Exception e) {
-                logger.error("Error calling global VMHook action: {}", e);
-            }
-        }
+        logger.warn("This deprecated method doesn't set VM hook. Define your hook component as a Spring bean.");
```
mkalinin: Let's preserve `setVmHook` logic to make it backward compatible.

ethereumj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -87,16 +89,17 @@
...
                                ProgramInvokeFactory programInvokeFactory, Block currentBlock) {
 
-        this(tx, coinbase, track, blockStore, programInvokeFactory, currentBlock, new EthereumListenerAdapter(), 0);
+        this(tx, coinbase, track, blockStore, programInvokeFactory, currentBlock, new EthereumListenerAdapter(), 0, VMHook.EMPTY);
     }
 
     public TransactionExecutor(Transaction tx, byte[] coinbase, Repository track, BlockStore blockStore,
                                ProgramInvokeFactory programInvokeFactory, Block currentBlock,
-                               EthereumListener listener, long gasUsedInTheBlock) {
+                               EthereumListener listener, long gasUsedInTheBlock, VMHook vmHook) {
```
zilm13: I'd keep old constructor in place, add new one for this and also add method `withVMHook`.
I guess `TransactionExecutor` is mostly used of all VM abstractions as one of the highest level and it's better to keep old interface unchanged.

PR #1182, opened 2018-09-20
VM hook redesign (optimized)

eugene-shevchenko:
Adds performance optimization on proxy calls to nested and deprecated VM hooks.
Adds VM hooks testing support via `StandaloneBlockchain`.
Increases SpringFramework version up to `4.3.19.RELEASE`.


ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1412,7 +1430,7 @@ public void play(Program program) {
     @Deprecated
     public static void setVmHook(VMHook vmHook) {
         logger.warn("VM.setVmHook(VMHook vmHook) is deprecated method. Define your hook component as a Spring bean.");
```
zilm13: I think we should add line that it will affect only new VM instances

eugene-shevchenko: What do you mean? Could you explain more detailed?

ethereumj-core/src/main/java/org/ethereum/vm/hook/RootVmHook.java
```diff
@@ -39,35 +37,40 @@
 
     private static final Logger logger = LoggerFactory.getLogger("VM");
 
-    private final List<VMHook> hooks;
+    private VMHook[] hooks;
 
     @Autowired
-    public RootVmHook(Optional<List<VMHook>> hooks) {
-        this.hooks = hooks.orElseGet(ArrayList::new);
+    public RootVmHook(Optional<VMHook[]> hooks) {
```
Nashatyrev: Not a blocker, but I would prefer using List instead of arrays as far as it's not critical. As soon as we added `hasHooks` flag this place is no more performance critical

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
zilm13: Ahh ok, it looks like a very old code and as it's going to be deprecated, not an isuue

PR #1183, opened 2018-09-21
Removed unused imports from sources

aldoborrero:
There are a lot of sources that the imports are not being actively used. This PR removes those unused imports to leave the code even more clean.


mkalinin:
Thanks for your efforts on getting rid of that. But we used to not make such changes aside from updates in logic. Any other contribution would be much appreciated, pick any issue you like from the list and double check with us before starting to work on it.

PR #1187, opened 2018-09-26
Issue-1184: no distinction between uint and int

jondoe1337:
During regression testing the `org.ethereum.util.StandaloneBlockchainTest.encodeTest1()` failed because the assumption on the expected value was wrong, imho:

The written SolidityContract has a setter-alike function and the value that is set is tested. However the value is an unsigned-int, so i doubt that the expected value `-1` was correct due to the wrong uint implementation, but feel free to clear my mind.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
+++ b/ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
@@ -58,7 +58,8 @@ public abstract class SolidityType {
     public static SolidityType getType(String typeName) {
         if (typeName.contains("[")) return ArrayType.getType(typeName);
         if ("bool".equals(typeName)) return new BoolType();
-        if (typeName.startsWith("int") || typeName.startsWith("uint")) return new IntType(typeName);
+        if (typeName.startsWith("int")) return new IntType(typeName);
+        if (typeName.startsWith("uint")) return new UnsignedIntType(typeName);
         if ("address".equals(typeName)) return new AddressType();
         if ("string".equals(typeName)) return new StringType();
         if ("bytes".equals(typeName)) return new BytesType();
@@ -356,21 +357,15 @@ public abstract class SolidityType {
             return ByteUtil.bigIntegerToBytes(bi, 20);
         }
     }
-
-    public static class IntType extends SolidityType {
-        public IntType(String name) {
+    
+    public static abstract class NumericType extends SolidityType {
+        public NumericType(String name)
+        {
             super(name);
         }
 
-        @Override
-        public String getCanonicalName() {
-            if (getName().equals("int")) return "int256";
-            if (getName().equals("uint")) return "uint256";
-            return super.getCanonicalName();
-        }
-
-        @Override
-        public byte[] encode(Object value) {
+        BigInteger encodeInternal(Object value)
+        {
             BigInteger bigInt;
 
             if (value instanceof String) {
@@ -393,14 +388,21 @@ public abstract class SolidityType {
             } else {
                 throw new RuntimeException("Invalid value for type '" + this + "': " + value + " (" + value.getClass() + ")");
             }
-            return encodeInt(bigInt);
+            return bigInt;
+        }
+    }
+
+    public static class IntType extends NumericType {
+        public IntType(String name) {
+            super(name);
         }
 
         @Override
-        public Object decode(byte[] encoded, int offset) {
-            return decodeInt(encoded, offset);
+        public String getCanonicalName() {
+            if (getName().equals("int")) return "int256";
+            return super.getCanonicalName();
         }
-
+        
         public static BigInteger decodeInt(byte[] encoded, int offset) {
             return new BigInteger(Arrays.copyOfRange(encoded, offset, offset + 32));
         }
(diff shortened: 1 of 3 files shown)


ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
```diff
@@ -356,21 +357,15 @@ public Object decode(byte[] encoded, int offset) {
             return ByteUtil.bigIntegerToBytes(bi, 20);
         }
     }
-
-    public static class IntType extends SolidityType {
-        public IntType(String name) {
+    
+    public static abstract class NumericType extends SolidityType {
+        public NumericType(String name)
+        {
```
zilm13: Could you, please, move opening bracket `{` to the method definition line like in surrounding code in this all following cases.

ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
```diff
@@ -393,23 +388,73 @@ public String getCanonicalName() {
...
         }
         public static byte[] encodeInt(BigInteger bigInt) {
             return ByteUtil.bigIntegerToBytesSigned(bigInt, 32);
         }
+
+        @Override
+        public byte[] encode(Object value)
+        {
+            BigInteger bigInt = encodeInternal(value);
+            return encodeInt(bigInt);
```
zilm13: I'm not sure but maybe we should throw an exception here if `bigInt` < 0. Here and in encodeInt methods.
@mkalinin what do you think?

mkalinin: That makes sense to me 👍

ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
```diff
@@ -410,6 +412,48 @@ public static BigInteger decodeInt(byte[] encoded, int offset) {
...
+        @Override
+        public String getCanonicalName() {
+            if (getName().equals("uint")) return "uint256";
+            return super.getCanonicalName();
+        }
+        public static BigInteger decodeInt(byte[] encoded, int offset) {
+            return new BigInteger(1, Arrays.copyOfRange(encoded, offset, offset + 32));
+        }
+        public static byte[] encodeInt(int i) {
+            return encodeInt(new BigInteger("" + i));
```
zilm13: One more check here for i >= 0

ethereumj-core/src/test/java/org/ethereum/solidity/SolidityTypeTest.java
```diff
@@ -0,0 +1,36 @@
+package org.ethereum.solidity;
```
zilm13: I mean this thing:
```
/*
 * Copyright (c) [2016] [ <ether.camp> ]
 * This file is part of the ethereumJ library.
 *
 * The ethereumJ library is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Lesser General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * The ethereumJ library is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public License
 * along with the ethereumJ library. If not, see <http://www.gnu.org/licenses/>.
 */
```

ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
```diff
@@ -356,21 +357,15 @@ public Object decode(byte[] encoded, int offset) {
...
-        public String getCanonicalName() {
-            if (getName().equals("int")) return "int256";
-            if (getName().equals("uint")) return "uint256";
-            return super.getCanonicalName();
-        }
-
-        @Override
-        public byte[] encode(Object value) {
+        BigInteger encodeInternal(Object value)
+        {
```
zilm13: bracket, please :)

ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
zilm13: Please, move opening bracket to the above line:
```
public NumericType(String name) {
```

ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
zilm13: ahhh, it will go through another one, my miss

ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
jondoe1337: Sorry. :(

ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
jondoe1337: This one isn't necessary as `encodeInt(int i)` uses `static encodeInt(BigInteger bigInt)` where I placed the check. Insert anyway?

zilm13:
@jondoe1337 cool, thank you!
Will merge after test finished

PR #1198, opened 2018-10-01
Update gradle command

ashishb:
Replace compile with implementation as recommended here:
https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation


mkalinin:
This change makes sense in case if we use `java-library` plugin

ashishb:
What's the plan to use that? Afaik, "compile" is [deprecated](https://stackoverflow.com/questions/44231886/android-gradle-4-0-compile-dependency-deprecated-how-does-new-api-configuration) in Gradle.

mkalinin:
You know, maybe I was too fast with the decision.

mkalinin:
Thx 👍

PR #1203, opened 2018-10-03
Improve JUMPDEST analysis for init code

mkalinin:


mkalinin:
Rejecting it by myself, cause this change is negligible in terms of attack vector reducing.

PR #1204, opened 2018-10-04
CREATE2 gas cost update

mkalinin:
### Includes:
- Latest update to [EIP-1014](https://eips.ethereum.org/EIPS/eip-1014) regarding `SHA3` gas cost addition
- a bit of refactor

### Reference implementation:
- https://github.com/ethereum/go-ethereum/pull/17812

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/vm/VM.java
+++ b/ethereumj-core/src/main/java/org/ethereum/vm/VM.java
@@ -41,6 +41,7 @@ import static org.ethereum.crypto.HashUtil.sha3;
 import static org.ethereum.util.ByteUtil.EMPTY_BYTE_ARRAY;
 import static org.ethereum.util.ByteUtil.toHexString;
 import static org.ethereum.vm.OpCode.*;
+import static org.ethereum.vm.VMUtils.getSizeInWords;
 
 /**
  * The Ethereum Virtual Machine (EVM) is responsible for initialization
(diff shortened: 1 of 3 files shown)


ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -319,8 +320,11 @@ else if (!currentValue.isZero() && newValue.isZero()) {
                     break;
                 case SHA3:
                     gasCost = gasCosts.getSHA3() + calcMemGas(gasCosts, oldMemSize, memNeeded(stack.peek(), stack.get(stack.size() - 2)), 0);
-                    DataWord size = stack.get(stack.size() - 2);
-                    long chunkUsed = (size.longValueSafe() + 31) / 32;
+                    long size = stack.get(stack.size() - 2).longValueSafe();
+                    if (size == Long.MAX_VALUE) {
```
Nashatyrev: What is this check for?

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -396,6 +400,12 @@ else if (!currentValue.isZero() && newValue.isZero()) {
                 case CREATE2:
                     gasCost = gasCosts.getCREATE() + calcMemGas(gasCosts, oldMemSize,
                             memNeeded(stack.get(stack.size() - 2), stack.get(stack.size() - 3)), 0);
+                    long codeSize = stack.get(stack.size() - 3).longValueSafe();
+                    if (codeSize == Long.MAX_VALUE) {
```
Nashatyrev: What is this check for?

ethereumj-core/src/main/java/org/ethereum/vm/VM.java
mkalinin: This check is similar to check of [mem gas calculation](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/vm/VM.java#L145). And looks like it's already done in a call to `calcMemGas` prior to this check. I'll remove it in favor of code readability.

PR #1209, opened 2018-10-08
Adds fatal exceptions handling.

eugene-shevchenko:


ethereumj-core/src/main/java/org/ethereum/manager/BlockLoader.java
```diff
@@ -125,7 +126,9 @@ private void blockWork(Block block) {
      * @return <code>true</code> if all blocks within all dumps have been successfully imported, <code>false</code> otherwise.
      */
     public boolean loadBlocks(Function<Path, DumpWalker> walkerFactory, Path... paths) {
-        Assert.notEmpty(paths, "There is nothing to import.");
+        if (ArrayUtils.isEmpty(paths)) {
+            throw new FatalEthereumException("There is nothing to import.");
```
zilm13: This doesn't look like fatal input to me.
It's just 0 input, 0 processed, everything fine.

eugene-shevchenko: Yeah, you are right !
I changed this logic. Now the method returns `false` in such case.
According to this I fixed client's code in `Start` and `Harmony`.

zilm13:
One test is failed:
```
org.ethereum.manager.BlockLoaderTest > testNoDumpsLoading FAILED
    java.lang.AssertionError: Expected exception: java.lang.IllegalArgumentException
Running test: Test testInconsistentDumpsLoading(org.ethereum.manager.BlockLoaderTest)
```

ethereumj-core/src/main/java/org/ethereum/exception/FatalEthereumException.java
```diff
@@ -0,0 +1,21 @@
+package org.ethereum.exception;
```
zilm13: So this one and its parent is not used anymore (and its parent)
Do we need to keep it?
And also copyrights are missed both for this one and its parent.

eugene-shevchenko: Yes, we don't use it anymore in this case, but it can be useful in the future to throw an exception that could shutdown an app.

mkalinin: I'd keep it simple and leave only `FatalBeanException` for the moment.

Issue #1191, opened 2018-09-26
Beacon attester


mkalinin:
### Abstract
Implement attestation logic and attester service.

### Requirements
Main source of requirements should be the [spec](https://notes.ethereum.org/SCIg8AH5SA-O4C1G1LYZHQ?view#Per-block-processing). Attestation is a process of signing off on a segment of canonical chain.
Attestation service is a service that triggers attestation logic in a specific slot that validator is assigned to, its functionality should be pretty similar to [ProposerService](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/main/java/org/ethereum/sharding/proposer/ProposerServiceImpl.java). There was a reasonable suggestion to use `AttestationRecord` for single attestation in the way it would be used for aggregated signature, with only one bit set in a bitfield and a signature that is aggregated with nothing.

~~`AttesterService` should produce an attestation after block with slot number that attester is assigned to has been imported. Attestation awaiting should be cancelled If block with that slot number does not occur in `2 * CYCLE_LENGTH` slots or its slot number becomes less than number of last finalized slot.~~

`AttesterService` should start producing an attestation at `Genesis.timestamp + slot * SLOT_DURATION + SLOT_DURATION / 2` where `slot` is a number of slot that attester is assigned to. And it should attest to the head of canonical chain, whatever block it will be.

One more thing that should be kept in mind is a way service is linked with committees, it's described in [Per block processing](https://notes.ethereum.org/SCIg8AH5SA-O4C1G1LYZHQ?view#Per-block-processing) section:
> Verify that the `parent.slot_number % len(get_shards_and_committees_for_slot(crystallized_state, parent.slot_number)[0].committee)`'th attester in `get_shards_and_committees_for_slot(crystallized_state, parent.slot_number)[0]` is part of the first (ie. item 0 in the array) `AttestationRecord` object; this attester can be considered to be the proposer of the parent block. In general, when a block is produced, it is broadcasted at the network layer along with the attestation from its proposer.

Basically, it reads about proposer only but other validators from mentioned committee must attest in that slot.

mkalinin:
Implemented in https://github.com/ethereum/ethereumj/pull/1210

PR #1210, opened 2018-10-09
Sharding/beacon attester

zilm13:
#1191 
not ready, just to talk

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
@@ -1,0 +1,39 @@
+package org.ethereum.sharding.crypto;
+
+import java.math.BigInteger;
+import java.util.Objects;
+
+/**
+ * Signature utilities
+ * Signature should be implemented using BLS
+ */
+public interface Sign {
+
+    Signature sign(byte[] msg, byte[] privateKey);
+
+    boolean verify(Signature signature, byte[] publicKey);
+
+    Signature aggSigns(Signature[] signatures);
+
+    class Signature {
+        public BigInteger r;
+        public BigInteger s;
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Signature signature = (Signature) o;
+            return Objects.equals(r, signature.r) &&
+                    Objects.equals(s, signature.s);
+        }
+
+        @Override
+        public String toString() {
+            return "Signature{" +
+                    "r=" + r +
+                    ", s=" + s +
+                    '}';
+        }
+    }
+}
--- a/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
@@ -1,6 +1,7 @@
 package org.ethereum.sharding.crypto;
 
 import java.math.BigInteger;
+import java.util.List;
 import java.util.Objects;
 
 /**
@@ -12,22 +13,27 @@ public interface Sign {
     /**
      * Sign the message
      */
-    Signature sign(byte[] msg, byte[] privateKey);
+    Signature sign(byte[] msg, BigInteger privateKey);
 
     /**
      * Verifies whether signature is made by signer with publicKey
      */
-    boolean verify(Signature signature, byte[] publicKey);
+    boolean verify(Signature signature, byte[] msg, BigInteger publicKey);
 
     /**
-     * Recovers public key using signature and hash of the message that was signed
+     * Calculates public key from private
      */
-    byte[] recover(Signature signature, byte[] msgHash);
+    BigInteger privToPub(BigInteger privKey);
 
     /**
      * Aggregates several signatures in one
      */
-    Signature aggSigns(Signature[] signatures);
+    Signature aggSigns(List<Signature> signatures);
+
+    /**
+     * Aggregates public keys
+     */
+    BigInteger aggPubs(List<BigInteger> pubKeys);
 
     class Signature {
         public BigInteger r;
--- a/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
@@ -9,10 +9,19 @@ import java.util.Objects;
  */
 public interface Sign {
 
+    /**
+     * Sign the message
+     */
     Signature sign(byte[] msg, byte[] privateKey);
 
+    /**
+     * Verifies whether signature is made by signer with publicKey
+     */
     boolean verify(Signature signature, byte[] publicKey);
 
+    /**
+     * Aggregates several signatures in one
+     */
     Signature aggSigns(Signature[] signatures);
 
     class Signature {
--- a/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
@@ -20,6 +20,11 @@ public interface Sign {
     boolean verify(Signature signature, byte[] publicKey);
 
     /**
+     * Recovers public key using signature and hash of the message that was signed
+     */
+    byte[] recover(Signature signature, byte[] msgHash);
+
+    /**
      * Aggregates several signatures in one
      */
     Signature aggSigns(Signature[] signatures);
--- a/ethereumj-core/src/main/java/org/ethereum/sharding/domain/BeaconGenesis.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sharding/domain/BeaconGenesis.java
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.io.ByteStreams;
+import org.ethereum.sharding.processing.state.AttestationRecord;
 import org.ethereum.util.Utils;
 
 import java.io.InputStream;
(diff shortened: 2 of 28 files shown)


ethereumj-core/src/main/java/org/ethereum/sharding/validator/BeaconAttesterImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        // validate that block doesn't contain our attestation
+        for (AttestationRecord attestationRecord : block.getAttestations()) {
+            if (Bitfield.hasVoted(attestationRecord.getAttesterBitfield(), index.getValidatorIdx())) {
+                throw new RuntimeException("Shouldn't attest block again");
+            }
+        }
+
+        BeaconState headState = repository.get(store.getCanonicalHead().getHash());
```
mkalinin: Look at the way of how state and head are passed to `BeaconProposer`. I think it should be the same for attester. At least it looks more robust.

ethereumj-core/src/main/java/org/ethereum/sharding/validator/BeaconAttesterImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        long lastJustified = headState.getCrystallizedState().getFinality().getLastJustifiedSlot();
+        AttestationRecord attestationRecord = new AttestationRecord(
+                slotNumber,
+                index.getShardId(),
+                new byte[0][0], // FIXME: obliqueParentHashes
+                new byte[32], // FIXME: shardBlockHash??
```
mkalinin: This is basically a hash of the block that is currently a head of canonical chain.

ethereumj-core/src/main/java/org/ethereum/sharding/validator/BeaconAttesterImpl.java
```diff
@@ -0,0 +1,78 @@
...
+                throw new RuntimeException("Shouldn't attest block again");
+            }
+        }
+
+        BeaconState headState = repository.get(store.getCanonicalHead().getHash());
+        long lastJustified = headState.getCrystallizedState().getFinality().getLastJustifiedSlot();
+        AttestationRecord attestationRecord = new AttestationRecord(
+                slotNumber,
+                index.getShardId(),
+                new byte[0][0], // FIXME: obliqueParentHashes
```
mkalinin: Basically, there is no need to fix it right here. It's always an empty array for single attestation

ethereumj-core/src/main/java/org/ethereum/sharding/validator/BeaconAttesterImpl.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public AttestationRecord attestBlock(long slotNumber, Committee.Index index, Beacon block, byte[] pubKey) {
+        if (block.getAttestations().length == 0) {
+            throw new RuntimeException("Block should have at least 1 attestation of its proposer");
+        }
+
+        // validate that block doesn't contain our attestation
+        for (AttestationRecord attestationRecord : block.getAttestations()) {
+            if (Bitfield.hasVoted(attestationRecord.getAttesterBitfield(), index.getValidatorIdx())) {
```
mkalinin: Hmm, it doesn't look like a possible case. Block contains attestations to previous blocks but it can not attest on itself.

ethereumj-core/src/main/java/org/ethereum/sharding/validator/BeaconProposerImpl.java
```diff
@@ -74,10 +81,25 @@ public BeaconProposerImpl(Randao randao, StateRepository repository,
     @Override
     public Beacon createNewBlock(Input in, byte[] pubKey) {
         Beacon block = new Beacon(in.parent.getHash(), randaoReveal(in.state, pubKey), in.mainChainRef,
-                HashUtil.EMPTY_DATA_HASH, in.slotNumber);
+                HashUtil.EMPTY_DATA_HASH, in.slotNumber, new AttestationRecord[0]);
         BeaconState newState = stateTransition.applyBlock(block, in.state);
         block.setStateHash(newState.getHash());
 
+        AttestationRecord[] attestationRecords = new AttestationRecord[1];
```
mkalinin: > Verify that an attestation from this validator is part of the first (ie. item 0 in the array) AttestationRecord object; this attester can be considered to be the proposer of the parent block. In general, when a block is produced, it is broadcasted at the network layer along with the attestation from its proposer.

I was totally wrong that proposer should _include_ this attestation to the block it proposes. As to the spec it should broadcast block and zero attestation alongside to the network. And its ancestor should include this attestation at the first position in the array of attestations. This totally makes sense, cause including zero attestation into the same block will inevitably change its hash and then the attestation won't pass verification part.

mkalinin: Also, zero attestation should be a part of `ValidatorService` rather than placed here

zilm13:
@mkalinin ready for review

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
```diff
@@ -0,0 +1,48 @@
+package org.ethereum.sharding.crypto;
+
+import java.math.BigInteger;
+import java.util.Objects;
+
+/**
+ * Signature utilities
+ * Signature should be implemented using BLS
+ */
+public interface Sign {
```
mkalinin: This interface lacks, at least, `recover` method that recovers a pub key from signature with usage of msg hash

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Verifies whether signature is made by signer with publicKey
+     */
+    boolean verify(Signature signature, byte[] publicKey);
+
+    /**
+     * Aggregates several signatures in one
+     */
+    Signature aggSigns(Signature[] signatures);
```
mkalinin: I'd move it to a separate interface? what do you think? The product of `aggSigns` is not of `Signature` type, there should be another model for that.

ethereumj-core/src/main/java/org/ethereum/sharding/domain/BeaconGenesis.java
```diff
@@ -53,7 +53,7 @@
     private List<String> initialValidators = new ArrayList<>();
 
     BeaconGenesis(Json json) {
-        super(json.parentHash(), json.randaoReveal(), json.mainChainRef(), EMPTY, SLOT);
+        super(json.parentHash(), json.randaoReveal(), json.mainChainRef(), EMPTY, SLOT, new ArrayList<>());
```
mkalinin: Maybe it's better to use `Collections.emptyList()` instead

ethereumj-core/src/main/java/org/ethereum/sharding/processing/state/ActiveState.java
```diff
@@ -0,0 +1,171 @@
...
+
+import static org.ethereum.crypto.HashUtil.blake2b;
+import static org.ethereum.sharding.processing.consensus.BeaconConstants.CYCLE_LENGTH;
+import static org.ethereum.util.ByteUtil.isSingleZero;
+import static org.ethereum.util.ByteUtil.toHexString;
+
+/**
+ * Active beacon chain state
+ */
+public class ActiveState {
```
mkalinin: If we introduce a new state it may be, also, reasonable to add a transition of that state

ethereumj-core/src/main/java/org/ethereum/sharding/processing/state/AttestationRecord.java
```diff
@@ -0,0 +1,169 @@
...
+ */
+public class AttestationRecord {
+
+    // Slot number
+    private final long slot;
+    // Shard ID
+    private final int shardId;
+    // List of block hashes that this signature is signing over that
+    // are NOT part of the current chain, in order of oldest to newest
+    private final byte[][] obliqueParentHashes;
```
mkalinin: Let's use a `List` container instead of array

ethereumj-core/src/main/java/org/ethereum/sharding/util/Bitfield.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.List;
+
+/**
+ * Bitfield is bit array where every bit represents status of
+ * attester with corresponding index
+ */
+public class Bitfield {
```
mkalinin: Looks like the only reasonable `static` method here is `createEmpty`. Would it be better to have the others as an instance methods?

ethereumj-core/src/main/java/org/ethereum/sharding/util/Bitfield.java
```diff
@@ -0,0 +1,132 @@
...
+        this.payload = BitSet.valueOf(data);
+    }
+
+    /**
+     * Calculates attesters bitfield length
+     * @param num  Number of attesters
+     * @return  Bitfield length in bytes
+     */
+    private static int calcLength(int num) {
+        return (num + 7) / Byte.SIZE;
```
mkalinin: Consider an overflow check here. @Nashatyrev proposed a good and simple solution for such cases https://github.com/ethereum/ethereumj/blob/a207aeef0acbeb74e8c7d2bc5d248e4079673eea/ethereumj-core/src/main/java/org/ethereum/vm/VMUtils.java#L160-L162

ethereumj-core/src/main/java/org/ethereum/sharding/validator/ValidatorService.java
```diff
@@ -51,7 +52,7 @@ default void propose(long slotNumber, int validatorIdx) {}
      * Submits a task to make an attestation in a given slot number.
      * Thread safe.
      */
-    default void attest(long slotNumber, int validatorIdx) {}
+    default void attest(long slotNumber, Committee.Index index) {}
```
mkalinin: Let's make same change for `propose` either.

ethereumj-core/src/main/java/org/ethereum/sharding/processing/state/ActiveState.java
zilm13: Also we don't have separate transition for `CrystallizedState` and it's done in `BeaconStateTransition`. I'd add `ActiveState` changes there too when it is needed as it's in conjunction with `CrystallizedState` changes and both are parts of `BeaconStateTransition`

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
zilm13: From the specs:
```
# BLS aggregate signature
    'aggregate_sig': ['int256']
```
and it looks like there are just 2 of such ints there, same way as for Signature

zilm13:
@mkalinin please, take a look

PR #1217, opened 2018-10-18
Fix/ropsten forks

zilm13:
Add correct fork indentity, as there are a lot of peers with incorrect fork in Ropsten and there is a big chance that peer will not get block choice from the correct fork if no restriction is applied


AndreiShybin:
52073f7

AndreiShybin:
3ae595f

Issue #1193, opened 2018-09-26
Verify attestations in blocks


mkalinin:
### Abstract
Verify block's attestation records on before the import.

### Proposed implementation
Implement a couple of new [ValidationRule](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/main/java/org/ethereum/sharding/processing/validation/ValidationRule.java) and add it to the [BeaconValidator](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/main/java/org/ethereum/sharding/processing/validation/BeaconValidator.java).
Rules should do following checks:
- Attestation record with `0` index contains distinct valid attestation that is created by proposer of that block
- Attestation records with indices `> 0` are valid records and were created by attesters from committee that is eligible to attest on block's slot number

Consider block as invalid if any of these two checks fails.

zilm13:
resolved with #1219

PR #1219, opened 2018-10-22
Sharding/propose attestations

### LINKED ISSUE #1192: Propose blocks with attestations
### Abstract
Feature blocks created by [BeaconProposer](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/main/java/org/ethereum/sharding/proposer/BeaconProposerImpl.java) with attestations.

### Requirements
`AttestationRecord` is defined by the [spec](https://notes.ethereum.org/SCIg8AH5SA-O4C1G1LYZHQ?view#Per-block-processing).
There should be a pool with pending attestations that holds distinct attestation records that are coming from attesters. The pool is updated either on newly came attestation record or on newly imported block, attestations that came within the block must be evicted from the pool. There should be a background task that constantly aggregates attestations with same chain segment.

Attestation aggregation consists of setting bits in a bitfield and aggregated signature calculation. Note, only attestations that refer to same chain segment can be aggregated. When pending attestation pool is updated, previous attestation aggregation task should be cancelled and superseded with a new one unless it's locked by proposer. Proposer locks the most recent aggregation task, waits until it's finished and then includes produced attestation records into a block it is working on.

zilm13:
Should resolve #1192, #1193 
not finished
need help


ethereumj-core/src/main/java/org/ethereum/sharding/processing/consensus/BeaconStateTransition.java
```diff
@@ -55,6 +63,15 @@ public BeaconStateTransition(StateTransition<Dynasty> dynastyTransition,
...
         CrystallizedState crystallized = to.getCrystallizedState();
+        ActiveState activeState = to.getActiveState();
+
+        List<AttestationRecord> mergedAttestations = new ArrayList<>();
+        mergedAttestations.addAll(activeState.getPendingAttestations());
+        mergedAttestations.addAll(block.getAttestations());
+        block.getAttestations().forEach(at -> beaconAttester.purgeAttestations(at));
+
+        activeState = activeState
+                .withPendingAttestations(mergedAttestations);
```
mkalinin: It's better to have something like `addAttestations(List<AttestationRecord> newAttestations)` to avoid the above merging.

ethereumj-core/src/main/java/org/ethereum/sharding/processing/consensus/BeaconStateTransition.java
```diff
@@ -55,6 +63,15 @@ public BeaconStateTransition(StateTransition<Dynasty> dynastyTransition,
     public BeaconState applyBlock(Beacon block, BeaconState to) {
 
         CrystallizedState crystallized = to.getCrystallizedState();
+        ActiveState activeState = to.getActiveState();
+
+        List<AttestationRecord> mergedAttestations = new ArrayList<>();
+        mergedAttestations.addAll(activeState.getPendingAttestations());
+        mergedAttestations.addAll(block.getAttestations());
+        block.getAttestations().forEach(at -> beaconAttester.purgeAttestations(at));
```
mkalinin: What do you think if we drive this part by a pub/sub mechanism? It would let us avoid tight components coupling.

ethereumj-core/src/main/java/org/ethereum/sharding/processing/consensus/BeaconStateTransition.java
```diff
@@ -67,8 +84,19 @@ public BeaconState applyBlock(Beacon block, BeaconState to) {
                     .withDynasty(dynasty)
                     .withLastStateRecalc(cycleStartSlot(block))
                     .withFinality(finality);
+
+            // remove attestations older than last_state_recalc
+            List<AttestationRecord> uptodateAttestations = new ArrayList<>();
```
mkalinin: I am starting to think of decoupling crystallized and active state transitions when looking at these lines.

ethereumj-core/src/main/java/org/ethereum/sharding/processing/consensus/BeaconStateTransition.java
```diff
@@ -67,8 +84,19 @@ public BeaconState applyBlock(Beacon block, BeaconState to) {
...
+            for (AttestationRecord record : activeState.getPendingAttestations()) {
+                if (record.getSlot() >= crystallized.getLastStateRecalc()) {
+                    uptodateAttestations.add(record);
+                }
+            }
+            beaconAttester.removeOldSlots(crystallized.getLastStateRecalc());
```
mkalinin: The same "pub/sub" note as previous one. I guess it could be even processed by same handler.

ethereumj-core/src/main/java/org/ethereum/sharding/processing/validation/ValidationResult.java
```diff
@@ -27,6 +27,7 @@
     Exist,
     NoParent,
     StateMismatch,
+    Invalid,
```
mkalinin: It should also be added to `ProcessingResult.fromValidation`. Looks like we need a note about that or better design to be sure that in future this problem won't arise again

ethereumj-core/src/main/java/org/ethereum/sharding/processing/BeaconChainImpl.java
```diff
@@ -120,9 +124,14 @@ public synchronized ProcessingResult insert(Beacon block) {
         if ((vRes = beaconValidator.validateAndLog(block)) != ValidationResult.Success)
             return ProcessingResult.fromValidation(vRes);
 
-        // apply state transition
         Beacon parent = pullParent(block);
         BeaconState currentState = pullState(parent);
+        if ((vRes = proposerValidator.validateAndLog(block, parent, currentState)) != ValidationResult.Success)
+            return ProcessingResult.fromValidation(vRes);
+        if ((vRes = attestationsValidator.validateAndLog(block, parent, currentState)) != ValidationResult.Success)
```
mkalinin: What if we merge `attestationsValidator` and `proposerValidator` into one validator with a couple of rules?

ethereumj-core/src/main/java/org/ethereum/sharding/processing/validation/ProposerValidator.java
```diff
@@ -0,0 +1,102 @@
...
+        rules.add((block, data) -> {
+            if (block.getSlotNumber() == 0) {
+                return Success;
+            }
+
+            CrystallizedState crystallized = data.state.getCrystallizedState();
+
+            Committee[][] committees = crystallized.getDynasty().getCommittees();
+
+            int index = (int) data.parent.getSlotNumber() % committees.length;
```
mkalinin: Look in `ValidatorServiceImpl` to see how this logic should look like. Maybe wrap it with some utility method to re-use?

ethereumj-core/src/main/java/org/ethereum/sharding/processing/validation/ProposerValidator.java
```diff
@@ -0,0 +1,102 @@
...
+
+            if (block.getAttestations().isEmpty()) {
+                return Invalid;
```
mkalinin: I'd increase the level of verbosity and call it `InvalidAttestations`

ethereumj-core/src/main/java/org/ethereum/sharding/processing/validation/ProposerValidator.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            AttestationRecord proposerAttestation = block.getAttestations().get(0);
+
+            if (proposerAttestation.getSlot() != data.parent.getSlotNumber()) {
```
mkalinin: It, also, should be checked that hash matched, shouldn't it?

PR #1221, opened 2018-10-25
Fix issue-1216: wrong signature generated for multidimensional arrays

jondoe1337:


Nashatyrev:
Thanks a lot for doing the fix PR!
I would avoid using `String.indexOf` tweaks where strict logic is applicable. 
What do you think about this fix variant https://github.com/ethereum/ethereumj/commit/baaccf744e804dc8be9808f7d6cd91f5046bfad0 ?

jondoe1337:
Looks better.

Nashatyrev:
You may change you PR then and I will merge

PR #1227, opened 2018-11-08
Improve ReceiptsDownloader: replace timeout with request counter - 1099

JonathanScialpi:
I reviewed issue 1099 and noticed the need to make the thread pause more efficient by removing the set 5 secs and replacing it with a dynamic wait time based on blocks to download.

I wanted to run some tests to verify that my change is working (besides the build) but noticed that all tests in ShortSync are failing unless I am doing something wrong. If so maybe someone can give me walkthrough on how to test properly?

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/sync/ReceiptsDownloader.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sync/ReceiptsDownloader.java
@@ -39,7 +39,6 @@ import org.springframework.stereotype.Component;
 
 import java.util.*;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static java.lang.Math.max;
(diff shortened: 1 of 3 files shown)


ethereumj-core/src/main/java/org/ethereum/sync/ReceiptsDownloader.java
```diff
@@ -167,12 +166,12 @@ private void retrieveLoop() {
         while (!Thread.currentThread().isInterrupted()) {
             try {
 
-                if (toDownload.isEmpty()) {
-                    if (fillBlockQueue() > 0 || System.currentTimeMillis() - t > REQUEST_TIMEOUT) {
-                        toDownload = getToDownload();
-                        t = System.currentTimeMillis();
-                    }
+              if (toDownload.isEmpty()) {
+                if (fillBlockQueue() > 0 || System.currentTimeMillis() - t > REQUEST_TIMEOUT || toDownload.size() == 0) {
```
mkalinin: `toDownload.size() == 0` is equal to `toDownload.isEmpty()` which is checked above, don't need to check it once again.

PR #1230, opened 2018-11-15
Solidity simple storage 예제를 실행할 수 있도록 StandaloneBlockchainSample을 수정.

kwanghoon:


zilm13:
This sample doesn't provide any new functionality usage examples comparing to `StandaloneBlockchainSample` https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/samples/StandaloneBlockchainSample.java .

PR #1236, opened 2018-11-26
Added Milagro crypto-library with BLS12-381 implementation. 

zilm13:
...First test of message signing added.
Not ready for merge yet, a lot of work to go.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
@@ -1,0 +1,121 @@
+package org.ethereum.sharding.crypto;
+
+import org.apache.milagro.amcl.BLS381.BIG;
+import org.apache.milagro.amcl.BLS381.ECP;
+import org.apache.milagro.amcl.BLS381.ECP2;
+import org.apache.milagro.amcl.BLS381.FP12;
+import org.apache.milagro.amcl.BLS381.PAIR;
+import org.apache.milagro.amcl.BLS381.ROM;
+import org.apache.milagro.amcl.RAND;
+
+import java.security.SecureRandom;
+
+import static org.apache.milagro.amcl.BLS381.BIG.MODBYTES;
+
+public class BLS381 {
+
+    public BIG CURVE_ORDER = new BIG(ROM.CURVE_Order);
+
+    public int PRIVATE_SIZE = MODBYTES;
+
+    public int ECP_POINT_SIZE = 2 * MODBYTES + 1;
+
+    public int ECP2_POINT_SIZE = 4 * MODBYTES;
+
+    private SecureRandom random = new SecureRandom();
+
+    /**
+     * Creates new random Signature (Private) key
+     */
+    public byte[] newSigKey() {
+        byte[] res = new byte[PRIVATE_SIZE];
+        RAND rand = new RAND();
+        byte[] randomBytes = new byte[PRIVATE_SIZE];
+        random.nextBytes(randomBytes);
+        rand.seed(PRIVATE_SIZE, randomBytes);
+        BIG randomNumber = BIG.randomnum(CURVE_ORDER, rand);
+        randomNumber.toBytes(res);
+
+        return res;
+    }
+
+    /**
+     * Obtains Verification (Public) key
+     * from Signature (Private) key on ECP2
+     */
+    public byte[] fromSigKey(BIG sigKey) {
+        ECP2 point = generator2();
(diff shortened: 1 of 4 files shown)


zilm13:
@mkalinin ready for review

zilm13:
resolves #1237

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
```diff
@@ -0,0 +1,53 @@
+package org.ethereum.sharding.crypto;
+
+import java.math.BigInteger;
+
+/**
+ * Common interface for BLS381-12 implementation
+ */
+public interface BLS381 {
+
+    BI generatePrivate();
```
mkalinin: I think naming is a bit incomplete cause it doesn't point that generation has _PRNG_ kind.

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
```diff
@@ -0,0 +1,53 @@
...
+
+    BI restorePrivate(BigInteger value);
```
mkalinin: I would abstract from any particular big number implementation. Wouldn't `byte[]` representation be sufficient enough?

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
```diff
@@ -0,0 +1,53 @@
...
+
+    BI restorePrivate(byte[] value);
```
mkalinin: It seems that this method addresses more general case than restoring just private keys. Maybe it's better to call it `restoreScalar`?

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
```diff
@@ -0,0 +1,53 @@
...
+
+    ECP2Point restoreECP2(byte[] value);
+
+    ECP2Point generator2();
+
+    ECP1Point mapToECP1(byte[] value);
+
+    FP12Point pair(ECP2Point pointECP2, ECP1Point pointECP1);
+
+    interface BI {
```
mkalinin: I'd call it `Scalar`

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
```diff
@@ -0,0 +1,53 @@
...
+        BigInteger asBigInteger();
```
mkalinin: Should be removed in case if we're not sticking with any specific big number implementations

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
```diff
@@ -0,0 +1,53 @@
...
+
+        byte[] asByteArray();
+    }
+
+    interface ECP1Point {
```
mkalinin: I think `ECP1` or even `P1` would be pretty enough here

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/BLS381.java
```diff
@@ -0,0 +1,53 @@
...
+        ECP2Point mul(BI value);
+
+        void add(ECP2Point value);
+
+        BigInteger asBigInteger();
+
+        byte[] asByteArray();
+    }
+
+    interface FP12Point {}
```
mkalinin: This is not a point, actually. Let's call it `FP12`. And I would, also, explicitly add `equals` method to that interface to point it out as an expecting one from underlying implementations

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
```diff
@@ -36,24 +44,30 @@
     BigInteger aggPubs(List<BigInteger> pubKeys);
 
     class Signature {
-        public BigInteger r;
-        public BigInteger s;
+        public BigInteger value;
```
mkalinin: It doesn't look to me that having `Signature` class is something useful. Maybe turn it into `byte[]`?

ethereumj-core/src/main/java/org/ethereum/sharding/crypto/Sign.java
```diff
@@ -36,24 +44,30 @@
     BigInteger aggPubs(List<BigInteger> pubKeys);
```
mkalinin: This method looks redundant. `pubKeys` may be passed to `verify` directly.

Issue #1235, opened 2018-11-23
Invocation fails with ArrayIndexOutOfBounds when multidimensional arrays are part of the signature


jondoe1337:
Referring issue #1216 - I reactivated my initial unittest and bumped into the next problem.
The function is now found by signature, but the inputdata cannot be decoded.

```
Thread [main] (Suspended (exception ArrayIndexOutOfBoundsException))	
	System.arraycopy(Object, int, Object, int, int) line: not available [native method]	
	Arrays.copyOfRange(byte[], int, int) line: 3521	
	SolidityType$IntType.decodeInt(byte[], int) line: 446	
	SolidityType$AddressType(SolidityType$IntType).decode(byte[], int) line: 456	
	SolidityType$AddressType.decode(byte[], int) line: 399	
	SolidityType$DynamicArrayType.decode(byte[], int) line: 273	
	SolidityType$StaticArrayType.decode(byte[], int) line: 204	
	SolidityType$StaticArrayType.decode(byte[], int) line: 163	
	CallTransaction$Function.decode(byte[], CallTransaction$Param[]) line: 158	
	CallTransaction$Function.decode(byte[]) line: 166	
	CallTransaction$Contract.parseInvocation(byte[]) line: 285
```

I created a new unittest, could need some help however:
```
@Test
    public void twoDimensionalArrayTypeAsParameter_isDecoded() {
        String funcJson = "{  \n" + 
                "      \"constant\":false,\n" + 
                "      \"inputs\":[  \n" + 
                "         {  \n" + 
                "            \"name\":\"orderAddresses\",\n" + 
                "            \"type\":\"address[5][]\"\n" + 
                "         },\n" + 
                "         {  \n" + 
                "            \"name\":\"orderValues\",\n" + 
                "            \"type\":\"uint256[6][]\"\n" + 
                "         },\n" + 
                "         {  \n" + 
                "            \"name\":\"fillTakerTokenAmounts\",\n" + 
                "            \"type\":\"uint256[]\"\n" + 
                "         },\n" + 
                "         {  \n" + 
                "            \"name\":\"v\",\n" + 
                "            \"type\":\"uint8[]\"\n" + 
                "         },\n" + 
                "         {  \n" + 
                "            \"name\":\"r\",\n" + 
                "            \"type\":\"bytes32[]\"\n" + 
                "         },\n" + 
                "         {  \n" + 
                "            \"name\":\"s\",\n" + 
[...]

PR #1239, opened 2018-11-27
Fix ABI code for multidimensional arrays

Nashatyrev:
* Fix for #1235 
* Fix encoding/decoding of fixed-size arrays with dynamic-size elements


ethereumj-core/src/main/java/org/ethereum/solidity/SolidityType.java
```diff
@@ -165,52 +180,40 @@ public SolidityType getElementType() {
...
-            return result;
+            return decodeTuple(encoded, offset, size);
         }
 
         @Override
         public int getFixedSize() {
-            // return negative if elementType is dynamic
-            return elementType.getFixedSize() * size;
+            if (isDynamicType()) {
+                return 32;
```
mkalinin: Where is this magic number from?

PR #1242, opened 2018-12-03
optimization p2p

lyy4j:


ethereumj-core/src/main/java/org/ethereum/net/rlpx/discover/NodeHandler.java
```diff
@@ -209,7 +210,10 @@ private void changeState(State newState) {
             sendPing();
         }
         state = newState;
-        stateChanged(oldState, newState);
+        if (newState == State.Active) {
+            stateChanged(oldState, newState);
+        }
```
Nashatyrev: I don't understand why this change doesn't break anything. Why `stateChanged()` should be called only on `Active`?

lyy4j: firstly, when the  oldest node challenge With the new node,and  the oldest node win,we should update the oldest  timestap,or else, the node will be get when  a new node was added and the bucket was full at next time.

secondly, as the origin code did ,when a node state was change ,the method stateChanged was called every time ,and the method is mainly to establish tcp connect with the peer node ,did it need ?

PR #1249, opened 2019-01-03
Fix #1248 and add test.

### LINKED ISSUE #1248: Transaction getKey is always null
```java
@Test
    public void testGetKeyFromSignature() {
        ECKey sender = new ECKey();
        BlockchainImpl blockchain = ImportLightTest.createBlockchain(GenesisLoader.loadGenesis(
                getClass().getResourceAsStream("/genesis/genesis-light.json")));
        Transaction tx = createTx(blockchain, sender, new byte[32], new byte[0], 100);
        Assert.assertNotNull(tx.getKey());
    }
```
tx.getKey() is always null.

jolestar:


ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
```diff
@@ -374,7 +374,7 @@ public boolean isContractCreation() {
 
     public ECKey getKey() {
         byte[] hash = getRawHash();
-        return ECKey.recoverFromSignature(signature.v, signature, hash);
+        return ECKey.recoverFromSignature(0, signature, hash);
```
mkalinin: Use `0` as `recId` will cause bugs. Take a look at [`ECKey.signatureToKeyBytes`](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/crypto/ECKey.java#L826) to get an idea of how `recId` is decoded from signature.

jolestar: use ECKey.signatureToKey(hash, getSignature())

PR #1255, opened 2019-01-23
DBSettings should be initialized in write lock.

chenjw13097:
Current code is:
```
    public void init(DbSettings settings) {
        (1) this.settings = settings;
        (2) resetDbLock.writeLock().lock();
        try {
            logger.debug("~> RocksDbDataSource.init(): " + name);
```

If two threads (named T1 and T2) call the method with different DbSettings, follow code may go wrong::
```
    options.setMaxOpenFiles(settings.getMaxOpenFiles());
    options.setIncreaseParallelism(settings.getMaxThreads());
```
For example, T1(1)->T2(1)->T1(2)->T2(2))，T1 will create the DB with settings passed by T2. For now, it still works well. But it may go wrong if init is called outside in future code.


zilm13:
@chenjw13097 if you don't want 2 threads to set different settings, you should either synchronize method or init settings only once, in constructor, I don't understand how could you prevent it with moving this line to `try` block

PR #1258, opened 2019-01-25
Add a version of constantinople HF without eip1283

mkalinin:
The numbers are taken from https://github.com/ethereum/go-ethereum/pull/18486


ethereumj-core/src/main/java/org/ethereum/config/net/RopstenNetConfig.java
```diff
@@ -29,5 +29,6 @@ public RopstenNetConfig() {
         add(10, new RopstenConfig(new HomesteadConfig()));
         add(1_700_000, new RopstenConfig(new ByzantiumConfig(new DaoHFConfig())));
         add(4_230_000, new RopstenConfig(new ConstantinopleConfig(new DaoHFConfig())));
+        add(4_939_394, new RopstenConfig(new PetersburgConfig(new DaoHFConfig())));
```
zilm13: Where could we check this number?

Issue #1186, opened 2018-09-26
Refactor TrieValidatorSet


mkalinin:
### What's wrong
Currently there are two sources in [TrieValidatorSet](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/main/java/org/ethereum/sharding/processing/db/TrieValidatorSet.java):
- Main source, a trie that stores `[validatorIndex: Validator]`
- Index, a key-value source `[pubKey: validatorIndex]`

According to potential use cases this schema doesn't look optimal.

### Proposal
- Switch trie to ~`[pubKey: Validator]`~ `[blake2b(pubKey): Validator]` schema, since in almost all use cases validator can be accessed by its `pubKey`.
- Switch Index to `[validatorIndex: pubKey]` which is useful for signature aggregation checks where validators are gonna be accessed by their indices. This seems to be the only case for the index.

PR #1259, opened 2019-01-31
Refactor TrieValidatorSet to close issue #1186

sunyaojin:
# Original Problem #1186 

Optimize [TrieValidatorSet](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/main/java/org/ethereum/sharding/processing/db/TrieValidatorSet.java). 

Currently, the two main sources are

`trie` -> `[validatorIndex: Validator]`
`index` -> a key-value source `[pubKey: validatorIndex]`

## Proposal of [#1186](https://github.com/ethereum/ethereumj/issues/1186)

`trie` -> `[blake2b(pubKey): validator]`
`index` -> a key-value source `[validatorIndex: pubKey]`

Main problem of this proposal is that, after the refactoring, when we use the function [getSnapshotTo](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/main/java/org/ethereum/sharding/processing/db/TrieValidatorSet.java#L135), the validator could not be found from source `index`. 

The main reason is that, `trie` got updated through `getSnapshotTo` but not the source of `index`. We can conclude that, `validatorIndex` is the data source we should also keep update. It means that, if we need to get access to validators through indices, the indices are always updated and correct. 

## New Proposal

As a result, I propose to make a small modification on two sources.

`trie` -> `[validatorIndex: pubKey]`
`index` -> `[blake2b(pubKey): validator]`

+ Two main sources can pass all the test on [TrieValidatorSetTest.java](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/test/java/org/ethereum/sharding/processing/db/TrieValidatorSetTest.java), including the [testSnapshotTo](https://github.com/ethereum/ethereumj/blob/research/sharding/ethereumj-core/src/test/java/org/ethereum/sharding/processing/db/TrieValidatorSetTest.java#L136).
+ `validatorIndex` is always updated; signature aggregation checks where validators are gonna be accessed by their indices can be processed correctly;
+ All the validators can be accessed by `pubKey`.

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/sharding/processing/db/TrieValidatorSet.java
+++ b/ethereumj-core/src/main/java/org/ethereum/sharding/processing/db/TrieValidatorSet.java
@@ -17,8 +17,8 @@
  */
 package org.ethereum.sharding.processing.db;
 
+import org.ethereum.crypto.HashUtil;
 import org.ethereum.datasource.NoDeleteSource;
-import org.ethereum.datasource.ObjectDataSource;
 import org.ethereum.datasource.Serializer;
 import org.ethereum.datasource.Source;
 import org.ethereum.datasource.SourceCodec;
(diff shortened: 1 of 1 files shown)


sunyaojin:
#1186

mkalinin:
We are not maintaining this implementation anymore. Sorry about that. Let me close all related issues to avoid any further confusion.

Issue #1171, opened 2018-09-06
NodeManager::setBootNodes should be public


matthewleon:
`org.ethereum.net.rlpx.discover.NodeManager::setBootNodes` should be `public` and not package-private. Without setting boot nodes, the object is very difficult to use outside of the package.

https://github.com/ethereum/ethereumj/blob/5c808c1ddfd743e668386d94ec68b6ba3f7713dd/ethereumj-core/src/main/java/org/ethereum/net/rlpx/discover/NodeManager.java#L120

netwalker2000:
I'v got this problem too. Without setting boot nodes, the object is very difficult to use outside of the package, considering the constructor is also public. I created a PR   https://github.com/ethereum/ethereumj/pull/1262

PR #1262, opened 2019-02-14
`org.ethereum.net.rlpx.discover.NodeManager::setBootNodes` should be …

netwalker2000:
…`public` and not package-private. Without setting boot nodes, the object is very difficult to use outside of the package.


netwalker2000:
from issue https://github.com/ethereum/ethereumj/issues/1171

mkalinin:
Thanks 👍

netwalker2000:
I'm hornored :)

PR #1057, opened 2018-04-22
Memory usage limited for outgoing eth responses

zilm13:
@mkalinin please, take a look

PR #1267, opened 2019-03-26
Null hash for transaction, mainly with Standalone blockchain

ESchouten:
Fix for empty array returned when (hash == null && parsed == true && rlpEncoded != null). See pull request #1057

### DIFF
--- a/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
@@ -473,7 +473,7 @@ public class Transaction {
         return rlpRaw;
     }
 
-    public byte[] getEncoded() {
+    public synchronized byte[] getEncoded() {
 
         if (rlpEncoded != null) return rlpEncoded;
 
--- a/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
+++ b/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java
@@ -256,8 +256,8 @@ public class Transaction {
     public byte[] getHash() {
         if (!isEmpty(hash)) return hash;
         rlpParse();
-        byte[] plainMsg = getEncoded();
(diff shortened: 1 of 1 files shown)


zilm13:
@ESchouten could you, please, add a test which is failing in current develop, but will pass with your fix. I couldn't reproduce the same bug and have not find the place in code where it could happen. Looks like I'm missing something. 
Also such test could be helpful not to open this bug again in future.

ESchouten:
@zilm13 I agree writing a unit test which demonstrates the race condition would be ideal. Problem is, I have not yet identified the source of this issue.
The error occurs while deploying a smart contract on the standalone blockchain on faster pc's and servers, while on slower servers and often in the IDE's debug mode it does not, making finding the source harder.
If it helps, we're using [https://github.com/adridadou/eth-propeller-ethj](eth-propeller)

ESchouten:
I am however able to repoduce the state of where rlpEncoded is initialized and the hash byte array is empty, though when trying to debug the origin of this, the race condition does not occur, probably due to the IDE pausing threads etc.

The only code I could find initializing the rlpEncoded array and not setting the hash value is located in InternalTransaction.java, though the class' method does not seem to be called.

zilm13:
@ESchouten so, it's race condition, in this case setting `getEncoded` to be `synchronized` makes more sense for me. What do you think?

ESchouten:
Agree, seems to work aswell! Still, in that case we're assuming hash string is set when `rlpEncoded` is set and `parsed` is true, which works right now, but might not be very maintainable.

E.g. The class InternalTransaction extends Transaction and overrides `getEncoded` but doesn't set the hash value.

zilm13:
@ESchouten why it's not maintainable?

`InternalTransaction` never goes out of VM

ESchouten:
@zilm13 Ah that explains.

ESchouten:
Reverted my change and made `getEncoded` synchronized to prevent the race condition, as you suggested.

PR #1273, opened 2019-04-03
Unit test fixes for solc-0.5.x for ABITest, CompilerTest, and Standal…

derekzuk:
Unit test fixes for solc-0.5.x for ABITest, CompilerTest, and StandaloneBlockchainTest.


derekzuk:
solcJ-all version needs to be changed in build.gradle

PR #1278, opened 2019-05-01
few optimizations on pruning

AbdelStark:
- add final modifier when appropriate for method parameters
- add final modifier when appropriate for class fields
- add final modifier when appropriate for class
- optimize `HashUtil.shortHash()`
- use `forEach` instead of loop to have more consistency in the class


AbdelStark:
@mkalinin can you review it please ?

PR #1281, opened 2019-08-07
Fix the comment.

feiqihang:
Fix the comment content according to the Yellow Paper.
nonce:
image
mixHash:
image


zilm13:
Thank you!

PR #1282, opened 2019-08-08
Fix MIN_GAS_LIMIT constant.

feiqihang:
Correct the MIN_GAS_LIMIT constant according to the Yellow Paper.
Otherwise, it may cause ethereumj to be incompatible with other language implementations.
image


zilm13:
@feiqihang It's not a mistake, minimum gas limit depends on network configuration. For Frontier and its children it's set to `5000`: https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/config/blockchain/FrontierConfig.java#L44-L47

PR #1283, opened 2019-08-08
Fix the validation rule for gasUsed.

feiqihang:
According to the agreement in the yellow book, it should be gasLimit <= gasUsed.
image


zilm13:
@feiqihang good catch! Thank you!

